PR #13, opened 2017-10-09
Amended content for session timeout and unauthorised views

miranda-hawkes:


conf/messages
```diff
@@ -1,3 +1,7 @@
-sessionTimeout.title = Session has timed out
+sessionTimeout.title = Your session has timed out
+sessionTimeout.helpOne = To manage your VAT account, you''ll have to
```
TomDudley92: Can this be changed `To view your VAT account ... `

PR #33, opened 2017-11-15
Added auth check to BTA-stub and base auth actions trait

lewishayqa:


app/controllers/BtaStubController.scala
```diff
@@ -22,16 +22,23 @@ import config.AppConfig
...
 import uk.gov.hmrc.play.bootstrap.controller.FrontendController
 
+import scala.concurrent.Future
+
 @Singleton
-class BtaStubController @Inject()(val messagesApi: MessagesApi,
-                                  btaStubService: BtaStubService,
-                                  implicit val appConfig: AppConfig) extends FrontendController with I18nSupport {
+class BtaStubController @Inject()(val messagesApi: MessagesApi, val authFunctions: AuthorisedFunctions,
+                                  val authConnector: AuthConnector, btaStubService: BtaStubService, implicit val appConfig: AppConfig)
```
miranda-hawkes: There is an `EnrolmentsAuthService` already that extends `AuthorisedFunctions` that you can inject instead and call `.authorised()` on

app/controllers/BtaStubController.scala
```diff
@@ -22,16 +22,23 @@ import config.AppConfig
...
-  val landingPage: Action[AnyContent] = Action.async { implicit request =>
-    btaStubService.getPartial().map { partial =>
-      Ok(views.html.btaStub.landingPage(partial))
+  def landingPage(): Action[AnyContent] = Action.async { implicit request =>
+    authFunctions.authorised() {
+      btaStubService.getPartial().map { partial =>
+        Ok(views.html.btaStub.landingPage(partial))
+      }
+    }.recoverWith {
+      case _: NoActiveSession => Future.successful(Unauthorized)
```
miranda-hawkes: If this returns `Unauthorized` it means it just shows a blank page which means we can't acceptance test hitting our 'dummy' BTA stubbed page and potentially should redirect to somewhere else - can discuss in person if that is needed at all - if not this is fine

PR #40, opened 2017-11-23
Added return obligation template and date formatters

miranda-hawkes:


test/views/templates/ReturnObligationTemplateSpec.scala
```diff
@@ -0,0 +1,94 @@
...
+import org.jsoup.Jsoup
+import org.jsoup.nodes.Document
+import views.ViewBaseSpec
+
+class ReturnObligationTemplateSpec extends ViewBaseSpec {
+
+  "returnObligation template" when {
+
+    object Selectors {
+      val p1 = "p:nth-of-type(1)"
```
macgrewal: Think about giving these selectors mor meaningful names e.g. `p2` could be renamed to `periodRangeDescription`

PR #42, opened 2017-11-24
BTAT-1174 VatApiConnector

WilliamJohnathonLea:


app/connectors/VatApiConnector.scala
```diff
@@ -0,0 +1,48 @@
...
+import models.Obligation.Status
+import models.Obligations
+import uk.gov.hmrc.http.HeaderCarrier
+import uk.gov.hmrc.play.bootstrap.http.HttpClient
+import utils.HttpResponseParsers.HttpGetResult
+
+import scala.concurrent.{ExecutionContext, Future}
+
+@Singleton
+class VatApiConnector @Inject()(http: HttpClient, appConfig: AppConfig) {
```
miranda-hawkes: Although you do most of the hard work in the `HttpResponseParsers` I think you still need a spec for `VatApiConnector`
i.e. a test for `obligationsUrl` and (maybe) a test for `getObligations()` using scalamock

app/connectors/VatApiConnector.scala
```diff
@@ -0,0 +1,48 @@
...
+
+  import utils.HttpResponseParsers.ObligationsReads
+
+  private[connectors] def obligationsUrl(vrn: String): String = s"${appConfig.vatApiBaseUrl}/vat/$vrn/obligations"
+
+  def getObligations(vrn: String, from: LocalDate, to: LocalDate, status: Status.Value)(implicit hc: HeaderCarrier, ec: ExecutionContext)
+  : Future[HttpGetResult[Obligations]] = {
+    val statusString = status match {
```
miranda-hawkes: Do you need this bit? Does `status.toString` not pull out the 'name'?

app/models/HttpError.scala
```diff
@@ -0,0 +1,56 @@
...
+ * limitations under the License.
+ */
+
+package models
+
+sealed trait HttpError {
+  def message: String
+}
+
+// === Obligation data errors ===
```
miranda-hawkes: Don't think we need this comment :)

app/utils/HttpResponseParsers.scala
```diff
@@ -0,0 +1,63 @@
...
+import play.api.http.Status._
+import play.api.libs.json.JsValue
+import uk.gov.hmrc.http.{HttpReads, HttpResponse}
+
+object HttpResponseParsers {
+
+  type HttpGetResult[T] = Either[HttpError, T]
+
+  implicit object ObligationsReads extends HttpReads[HttpGetResult[Obligations]] {
+    override def read(method: String, url: String, response: HttpResponse): HttpGetResult[Obligations] = {
```
miranda-hawkes: Can you add some logging here please? Logging for a `BAD_REQUEST` may be better done in `generateClientError()` so you can capture the specific error code

conf/application.conf
```diff
@@ -67,6 +67,10 @@ microservice {
         host = localhost
         port = 9250
       }
+      vat-api {
```
miranda-hawkes: Is there an app-config PR for this too?

app/connectors/VatApiConnector.scala
WilliamJohnathonLea: turns out it does, was just being safe

conf/application.conf
WilliamJohnathonLea: not yet, literally just made this change

app/utils/HttpResponseParsers.scala
WilliamJohnathonLea: The logging should go in the connecter not parser, because there would be a mix of concerns I put it in the parser. The parser is just meant to convert the response from the connector.

PR #58, opened 2017-12-13
BTAT-1268 Added partial connector and integration test

lewishayqa:


it/helpers/IntegrationBaseSpec.scala
```diff
@@ -38,7 +38,9 @@ trait IntegrationBaseSpec extends UnitSpec with WireMockHelper with GuiceOneServ
     "microservice.services.auth.host" -> mockHost,
     "microservice.services.auth.port" -> mockPort,
     "microservice.services.vat-api.host" -> mockHost,
-    "microservice.services.vat-api.port" -> mockPort
+    "microservice.services.vat-api.port" -> mockPort,
```
ja-green: Can you remove org.scalatest.TestSuite from the imports as it is not used

PR #83, opened 2018-01-19
Separated VAT details models into view & service/connector models

miranda-hawkes:
Also added some packages for payments & obligations models


app/models/viewModels/VatDetailsViewModel.scala
```diff
@@ -0,0 +1,29 @@
...
+package models.viewModels
+
+import java.time.LocalDate
+
+import play.api.libs.json.{Format, Json}
+
+case class VatDetailsViewModel(nextPaymentDueDate: Option[LocalDate],
+                               nextObligationDueDate: Option[LocalDate])
+
+object VatDetailsViewModel {
```
lewishayqa: Do you need a companion object and implicit JSON formatter for a view model?

PR #88, opened 2018-01-22
BTAT-1522: Added Account Details page

ja-green:


app/controllers/AccountDetailsController.scala
```diff
@@ -0,0 +1,58 @@
...
+import scala.concurrent.Future
+
+class AccountDetailsController @Inject()(val messagesApi: MessagesApi,
+                                         val enrolmentsAuthService: EnrolmentsAuthService,
+                                         implicit val appConfig: AppConfig)
+  extends AuthorisedController with I18nSupport {
+
+  def accountDetails(): Action[AnyContent] = authorisedAction { implicit request =>
+    user =>
+      for {
```
miranda-hawkes: In scala, `for {...} yield {...}` is useful when handling multiple futures, but if you're only needing to use one then using a `.map` is probably more appropriate
E.g.
`handleAccountDetailsModel(user).map{accountModel => 
   Ok(views.html.account.accountDetails(user, accountModel)
}`

test/controllers/AccountDetailsControllerSpec.scala
```diff
@@ -0,0 +1,144 @@
...
+          )
+        }
+
+        private val result = await(target.handleAccountDetailsModel(testUser))
+        result shouldBe exampleAccountDetailsModel
+      }
+    }
+  }
+}
+
```
miranda-hawkes: Can you remove the empty space at the bottom of this file please?

test/views/account/AccountDetailsViewSpec.scala
```diff
@@ -0,0 +1,106 @@
...
+    val correspondenceContactRow = "dt#correspondence-contact"
+    val correspondenceEmailRow = "dt#correspondence-email"
+    val correspondencePhoneRow = "dt#correspondence-phone"
+    val correspondenceMobileRow = "dt#correspondence-mobile"
+  }
+
+  private val user = User("1111")
+  val account = AccountDetailsModel("", "", "", "", "", "", "", "", "")
+
+  "Rendering the VAT details page" should {
```
miranda-hawkes: I think you need to add tests for the values of each row too to assert that you are rendering the model values correctly

PR #99, opened 2018-02-02
BTAT-1698 Stripped out the unused BTA header partial

lewishayqa:


app/views/main_template.scala.html
```diff
@@ -20,17 +20,11 @@
...
         appConfig: config.AppConfig)(mainContent: Html)(implicit request : Request[_], messages: Messages)
 
 @import uk.gov.hmrc.play.views.html.layouts
 
-@serviceInfoContent = {
-  @if(showBtaHeader) {
-    @serviceInfo
-  }
-}
+@serviceInfoContent = {}
```
TomDudley92: This can be removed as it is defaulted in govuk_wrapper.scala.html

PR #100, opened 2018-02-02
Added navigation breadcrumb template

miranda-hawkes:


miranda-hawkes:
Doing as part of BTAT-1668

PR #144, opened 2018-03-02
BTAT-1985: Added test-only route, controller and view for OPS stub

macgrewal:


app/testOnly/controllers/PaymentsStubController.scala
```diff
@@ -0,0 +1,47 @@
...
+class PaymentsStubController @Inject()(val messagesApi: MessagesApi,
+                                       val authConnector: AuthConnector,
+                                       authFunctions: AuthorisedFunctions,
+                                       implicit val appConfig: AppConfig)
+  extends FrontendController with I18nSupport {
+
+  def stub(): Action[AnyContent] = Action.async { implicit request =>
+    authFunctions.authorised() {
+      val data: String = request.session("payment-data")
+      val model: PaymentDetailsModel = Json.parse(data).as
```
lewishayqa: This line is causing an error for me and preventing compilation.
Error message: "diverging implicit expansion for type play.api.libs.json.Reads[V] 
starting with object StringReads in trait DefaultReads"

macgrewal: Sorry checked in the line I was tinkering with, update on the way

PR #147, opened 2018-03-02
BTAT-1993: Updated views to use payment feature switch

ja-green:


macgrewal:
@ja-green Can you change the name of the feature from payments to allowPayments fist please because the code as is is hard to understand without the verb. This will also effect #148

app/views/payments/openPayments.scala.html
```diff
@@ -72,15 +76,16 @@ <h1 class="heading-xlarge">@messages("openPayments.heading")</h1>
...
-      <span id="direct-debits" class="heading-small">Direct debits</span>
     </div>
 
-    <div id="check-direct-debit" class="form-group">
-      <p>
-        @Messages("openPayments.directDebitLink") <a href="#">@Messages("openPayments.directDebitLinkText")
-      </a> @Messages("openPayments.directDebitNotSure")</p>
-
-    </div>
+    @if(appConfig.features.allowPayments()) {
```
macgrewal: The allowPayments feature is seperate to the DD feature so it shouldn't stop users from going to DD

TomDudley92:
@ja-green Can you just check that this can be closed? I think this has been done under BTAT-1993-clean but can you just double check for me

PR #166, opened 2018-03-15
BTAT-2154: Updated content for  when no payment / return due

ja-green:


miranda-hawkes:
There are merge conflicts here. Also, is there a reason this was done before other higher-priority stories?

PR #173, opened 2018-03-19
Updated content for unauthorised page

miranda-hawkes:
Also removed sign-out button on session-timeout page
Note: requires an app-config-base PR merged too


app/config/frontendAppConfig.scala
```diff
@@ -58,6 +58,8 @@ trait AppConfig extends ServicesConfig {
   val surveyUrl: String
   val surveyThankYouUrl: String
   val signOutUrl: String
+  val mtdVatSignUpUrl: String
+  val unauthorisedSignOutUrl: String
```
lewishayqa: Unused import in this file.

PR #179, opened 2018-03-22
Logging fix

HarryGull:


app/connectors/httpParsers/ResponseHttpParsers.scala
```diff
@@ -33,8 +34,10 @@ trait ResponseHttpParsers {
   private def generateClientError(error: ApiError): Left[HttpError, Nothing] = {
     error match {
       case ApiSingleError(code, message) => Left(BadRequestError(code, message))
-      case ApiMultiError(_, _, _) => Left(MultipleErrors)
-      case ApiMultiErrorFinancial(_) => Left(MultipleErrors)
+      case ApiMultiError(code, message, errorResponse) =>
```
lewishayqa: Same issue here as in the parallel PR on view-vat-returns.

app/models/errors/httpErrors.scala
```diff
@@ -35,8 +35,9 @@ case class UnexpectedStatusError(code: Int, errorResponse: String) extends HttpE
     s"[UnexpectedStatusError]- RESPONSE status: $code, body: $errorResponse"
 }
 
-case object MultipleErrors extends HttpError {
-  override val message: String = "Received multiple errors."
+case class MultipleErrors(code: String, errorResponse: String) extends HttpError {
```
lewishayqa: Same question here as in your parallel PR on view-vat-returns regarding error _code_ type consistency.

PR #181, opened 2018-03-23
BTAT-2192 Updated vat overview with accessibility suggestions

lewishayqa:


app/views/vatDetails/details.scala.html
```diff
@@ -27,30 +27,44 @@
     currentPage = messages("vatDetails.title")
   )
 
-  <div class="grid-row">
+  <div class="grid-row form-group">
```
miranda-hawkes: This shouldn't really be used here as there are no inputs: http://govuk-elements.herokuapp.com/form-elements/
Are there alternate classes that could add padding?

PR #187, opened 2018-03-29
BTAT-2207 Raising coverage on vat-summary-frontend

WilliamJohnathonLea:


test/controllers/SurveyControllerSpec.scala
```diff
@@ -89,4 +89,55 @@ class SurveyControllerSpec extends ControllerBaseSpec {
 
   }
 
+  "posting invalid survey data" should {
+    "return a bad request" in new SurveyControllerTest {
+
+      val testModel: ExitSurveyAuditModel = ExitSurveyAuditModel(SurveyJourneyModel(Some("yes"),
```
lewishayqa: The vals _testModel_, _expectedData_ and the _sendEvent_ mocked function are the same in these three unit tests. Could you put these in the trait or elsewhere outside of the tests?

PR #235, opened 2018-05-25
BTAT-2769: Added payment history section to hub page

chrisjwwalker:


app/views/vatDetails/details.scala.html
```diff
@@ -18,56 +18,67 @@
...
-          detailsViewModel.paymentError
-        )
-        @views.html.templates.nextReturnSection(
-          detailsViewModel.nextReturnObligationDueDate,
-          detailsViewModel.returnObligationOverdue,
-          detailsViewModel.returnObligationError
-        )
-      </div>
+            <div class="grid-row">
+                @if(paymentHistoryEnabled) {
```
TomDudley92: Instead of using the bool passed in from the controller you can use the appConfig and then overwrite it in the tests to test it appearing in the view

app/controllers/VatDetailsController.scala
```diff
@@ -22,6 +22,7 @@ import audit.AuditingService
 import audit.models.{ViewNextOpenVatObligationAuditModel, ViewNextOutstandingVatPaymentAuditModel}
 import javax.inject.{Inject, Singleton}
 import config.AppConfig
+import config.features.Features
```
TomDudley92: Unused import

test/views/vatDetails/VatDetailsViewSpec.scala
```diff
@@ -156,6 +162,104 @@ class VatDetailsViewSpec extends ViewBaseSpec {
...
+
+      "have the text" in {
+        submittedReturns.select("p").text() shouldBe "Check the payments you've made."
+      }
+    }
+  }
+
+  "Rendering the VAT details page without the payment history section" should {
+
+    mockConfig.features.allowPaymentHistory(false)
```
TomDudley92: This can be removed due to it being set on line 253

PR #254, opened 2018-06-14
BTAT-2980: Added DirectDebitController

ja-green:


TomDudley92:
The OpenPaymentsController needs to call the getDirectDebitStatus method from the PaymentsService as this bool will need to be passed down to the view. A good example of how this has been done before is in the VatDetailsController

app/controllers/DirectDebitController.scala
```diff
@@ -58,7 +60,7 @@ class DirectDebitController @Inject()(val messagesApi: MessagesApi,
               Redirect(url)
             }
 
-          case Left(_) => InternalServerError(directDebitError())
-        }
+          case Left(_) => InternalServerError(standardError(appConfig, "Direct Debit Error", "Error", ""))
```
TomDudley92: This should be using the messages file as it will not translate to Welsh when eventually do this. Also the error message is incorrect. It should be Sorry, there is a problem with the service. Try again later. The key for this is `vatDetails.error`

app/controllers/OpenPaymentsController.scala
```diff
@@ -38,27 +38,40 @@ extends AuthorisedController with I18nSupport {
...
-        case Right(Some(payments)) =>
-          val model = getModel(payments.financialTransactions)
-          auditEvent(user, model)
-          Ok(views.html.payments.openPayments(user, model))
-        case Right(_) =>
-          auditEvent(user, Seq.empty)
-          Ok(views.html.payments.noPayments(user))
-        case Left(_) => InternalServerError(views.html.errors.paymentsError())
+
+      def something(hasActiveDirectDebit: Option[Boolean]) = {
```
TomDudley92: Please rename this method from something. Something is not a meaningful name

app/controllers/OpenPaymentsController.scala
```diff
@@ -38,27 +38,40 @@ extends AuthorisedController with I18nSupport {
...
+          case Right(_) =>
+            auditEvent(user, Seq.empty)
+            Ok(views.html.payments.noPayments(user))
+          case Left(_) => InternalServerError(views.html.errors.paymentsError())
+        }
+      }
+
+      paymentsService.getDirectDebitStatus(user.vrn).flatMap {
+        case Right(status) => something(Some(status))
+        case Left(_) => something(None)
```
TomDudley92: Can you add a test for the please

app/controllers/DirectDebitController.scala
```diff
@@ -24,7 +24,9 @@ import models.DirectDebitDetailsModel
 import play.api.i18n.{I18nSupport, MessagesApi}
 import play.api.mvc.{Action, AnyContent}
 import services.{EnrolmentsAuthService, PaymentsService}
-import views.html.errors.directDebitError
+import views.html.errors.standardError
+
+import scala.concurrent.Future
```
TomDudley92: Can you remove the unused import please

app/models/viewModels/OpenPaymentsViewModel.scala
```diff
@@ -12,21 +12,11 @@
...
- *@
+ */
 
-@()(implicit appConfig: config.AppConfig, request: Request[_], messages: Messages)
+package models.viewModels
 
-@contentHeader = {
-  <h1>@messages("paymentsError.heading")</h1>
-}
+import java.time.LocalDate
```
TomDudley92: Please remove the unused import

test/audit/DirectDebitAuditModelSpec.scala
```diff
@@ -0,0 +1,91 @@
...
+ * limitations under the License.
+ */
+
+package audit
+
+import java.time.LocalDate
+
+import audit.models.DirectDebitAuditModel
+import _root_.models.DirectDebitDetailsModel
+import _root_.models.obligations.{VatReturnObligation, VatReturnObligations}
```
TomDudley92: Can you remove the unused imports please

PR #258, opened 2018-06-18
Added direct debit stub page and controller

daithi-o-gorduin:


HarryGull:
This page should display the posted JSON values from the handoff so we can see it is woking as expected.
it should then extract the returnUrl and navigate to that when a continue button or similar is clicked..

HarryGull:
Update.

To work properly.
The connector should be configured to go to the dummy controller with same url as direct debits would be.

The dummy controller should then validate the JSON in code (there is a boby.validate method that can do this). If valid the controlelr should then send a reponse back with a nextUrl in the body.
This next URL should be to go to our dummy test page.
We can't load the JSON values here but we already validated them previously.

The page should then have a next/contine link/button that can be hard coded to go back to the same place as the returnUrl initially posted.

The above simulates the full payments handoff without actually going to payments.
e.g. the nextURl extraction etc.

PR #272, opened 2018-06-22
BTAT-3118 Added new update vat details section on vat details view

TomDudley92:


test/views/vatDetails/VatDetailsViewSpec.scala
```diff
@@ -136,16 +137,18 @@ class VatDetailsViewSpec extends ViewBaseSpec with BeforeAndAfterEach {
       elementText(Selectors.vatRegNo) shouldBe s"VAT registration number (VRN): ${user.vrn}"
     }
 
-    "have the account details section" should {
+    "have the vat certificate section" should {
```
lewishayqa: I think this ViewSpec also needs tidying up in general. There seems to be some inconsistencies and repeated tests. For example, I wouldn't have thought we would need to repeat the all of the tests for the scenario where we hide the payment history section.

PR #276, opened 2018-06-27
BTAT-3138: Added payment history helper

ja-green:


app/views/templates/paymentsHistoryCharge.scala.html
```diff
@@ -0,0 +1,77 @@
...
+  (firstDate, secondDate) match {
+    case (Some(fd), Some(sd)) => {
+      messages(
+        s"$chargeTypeDescription",
+        s"${fd.getDayOfMonth} ${fd.getMonth.toString.head}${fd.getMonth.toString.toLowerCase.substring(1, 3)}",
+        s"${sd.getDayOfMonth} ${sd.getMonth.toString.head}${sd.getMonth.toString.toLowerCase.substring(1, 3)} ${sd.getYear}"
+      )
+    }
+
+    case (Some(fd), None) => {
```
TomDudley92: To increase the coverage can you add tests for the first or second date being a None please. I'm not sure though that we can get a charge with only one date though so those cases may not be required

PR #279, opened 2018-06-29
BTAT-3139 Create charge type helper for what-you-owe page

lewishayqa:


TomDudley92:
Can you update the hidden context please? 

`£999 overdue a VAT officer’s investigation showed you underpaid by this amount` This doesn't make too much sense. I spoke to Kate and the content should be `£999 is overdue, a VAT officer’s investigation showed you underpaid by this amount`

PR #295, opened 2018-07-19
BTAT-3377: Changed table headers to be hidden

ja-green:


app/views/payments/paymentHistory.scala.html
```diff
@@ -45,13 +45,22 @@ <h2>@model.selectedYear.toString</h2>
...
         <p>@messages("paymentsHistory.noHistory", model.selectedYear.toString)</p>
+
       } else {
         <table>
           <thead>
             <tr>
-              <th scope="col">@messages("paymentsHistory.paymentReceivedTitle")</th>
-              <th scope="col">@messages("paymentsHistory.descriptionTitle")</th>
-              <th class="numeric" scope="col">@messages("paymentsHistory.amountTitle")</th>
+              <th scope="col">
```
lewishayqa: Could you update the table heading tests in some way because of this change? Perhaps rewording the test descriptions and updating the CSS selectors to use `visuallyhidden` just for sanity that they are in fact being hidden?

app/views/payments/paymentHistory.scala.html
```diff
@@ -45,13 +45,22 @@ <h2>@model.selectedYear.toString</h2>
```
lewishayqa: I know this isn't a part of your subtask, but while you're here could you set the `allowPaymentHistory` feature to true in application.conf please. It's set to true in app-config-base and in Production so it should be true locally too.

PR #307, opened 2018-07-26
BTAT-3412: Removed unused features

ja-green:


test/config/features/FeatureSpec.scala
```diff
@@ -27,18 +27,84 @@ class FeatureSpec extends PlaySpec with GuiceOneAppPerSuite with BeforeAndAfterE
...
+
+    features.userResearchBanner(true)
+    features.allowDirectDebits(true)
+    features.staticDateEnabled(true)
+    features.vatCertificateEnabled(true)
+    features.enableVatObligationsService(false)
+    features.useDirectDebitDummyPage(true)
+  }
+
+  "The User Research Banner Feature" should {
```
lewishayqa: I think just testing a single feature should be enough here as it looks like this is testing the .apply() function multiple times.

test/controllers/PaymentHistoryControllerSpec.scala
```diff
@@ -114,8 +114,6 @@ class PaymentHistoryControllerSpec extends ControllerBaseSpec {
     val mockDateService: DateService = mock[DateService]
```
lewishayqa: Unused import in this file.

app/controllers/AuthorisedController.scala
```diff
@@ -38,16 +38,12 @@ abstract class AuthorisedController extends FrontendController with I18nSupport
   def authorisedAction(block: Request[AnyContent] => User => Future[Result]): Action[AnyContent] = Action.async {
```
lewishayqa: Unused import in this file.

PR #319, opened 2018-09-27
BTAT-3981: Updated plugins for new build

ja-green:


build.sbt
```diff
@@ -54,7 +55,7 @@ lazy val coverageSettings: Seq[Setting[_]] = {
 
 val compile: Seq[ModuleID] = Seq(
   ws,
-  "uk.gov.hmrc" %% "bootstrap-play-25" % "3.4.0",
+  "uk.gov.hmrc" %% "bootstrap-play-25" % "3.8.0",
```
lewishayqa: I left a message in Slack about this yesterday, but basically - don't update this dependency now. It changes the way `ExtendedDataEvent` works and will change our auditing. We have a story on the backlog to address this in the future but for now just keep it as it was.

PR #334, opened 2018-12-05
Btat 4535

mitchell1972:


app/connectors/httpParsers/PaymentsHistoryHttpParser.scala
```diff
@@ -50,7 +50,8 @@ object PaymentsHistoryHttpParser extends ResponseHttpParsers {
       FinancialTransactionsConstants.vatDefaultSurcharge,
       FinancialTransactionsConstants.errorCorrectionCreditCharge,
       FinancialTransactionsConstants.errorCorrectionDebitCharge,
-      FinancialTransactionsConstants.vatRepaymentSupplement
+      FinancialTransactionsConstants.vatRepaymentSupplement,
+      FinancialTransactionsConstants.vatAdditionalAssessmentInterest
```
LadonJackson: I don't think we are doing the chargeTypes for the payment history as a part of this story, just the 'what you owe' page stuff, so I think this should be done in the PaymentHttpParser

app/models/viewModels/PaymentsHistoryModel.scala
```diff
@@ -47,7 +47,8 @@ object PaymentsHistoryModel {
         FinancialTransactionsConstants.vatCentralAssessment,
         FinancialTransactionsConstants.vatDefaultSurcharge,
         FinancialTransactionsConstants.errorCorrectionChargeType,
-        FinancialTransactionsConstants.vatRepaySupplement
+        FinancialTransactionsConstants.vatRepaySupplement,
+        FinancialTransactionsConstants.vatAdditionalAssessmentInterest
```
LadonJackson: I don't think we need to do this if we aren't doing the payment history yet

app/views/templates/PaymentsHistoryChargeHelper.scala
```diff
@@ -70,9 +70,16 @@ object PaymentsHistoryChargeHelper {
...
+     FinancialTransactionsConstants.vatAdditionalAssessmentInterest,
+    "paymentsHistory.AADefaultInterestTitle",
+    "paymentsHistory.AADefaultInterestDescription"
+
+  )
+
   val values = Seq(VatReturnDebitCharge, VatReturnCreditCharge, VatOfficerAssessmentCreditCharge,
     VatOfficerAssessmentDebitCharge, VatCentralAssessment, VatDefaultSurcharge,
-    ErrorCorrectionDebitCharge, ErrorCorrectionCreditCharge, VatRepaymentSupplement)
+    ErrorCorrectionDebitCharge, ErrorCorrectionCreditCharge, VatRepaymentSupplement,AADefaultInterest)
```
LadonJackson: I don't think we need to do these yet either

conf/messages
```diff
@@ -132,6 +132,8 @@ paymentsHistory.vatErrorCorrectionCreditChargeTitle =  Error correction repaymen
 paymentsHistory.vatErrorCorrectionChargeDescription = for correcting your {0} return
 paymentsHistory.vatRepaymentSupplementTitle = Late repayment compensation from HMRC
 paymentsHistory.vatRepaymentSupplementDescription = we took too long to repay your {0} return
+paymentsHistory.AADefaultInterestDescription = interest charged on additional tax assessed for the period {0} to {0}
+paymentsHistory.AADefaultInterestTitle = Additional assessment interest
```
LadonJackson: I don't think this is actually used anywhere.

conf/messages
```diff
@@ -132,6 +132,8 @@ paymentsHistory.vatErrorCorrectionCreditChargeTitle =  Error correction repaymen
 paymentsHistory.vatErrorCorrectionChargeDescription = for correcting your {0} return
 paymentsHistory.vatRepaymentSupplementTitle = Late repayment compensation from HMRC
 paymentsHistory.vatRepaymentSupplementDescription = we took too long to repay your {0} return
+paymentsHistory.AADefaultInterestDescription = interest charged on additional tax assessed for the period {0} to {0}
```
LadonJackson: I think this should be
`interest charged on additional tax assessed for the period {0}`
because `displayDateRange` is given to the message which sorts out the start and end date for you

PR #359, opened 2019-01-23
BTAT-4897

TomDudley92:
Added several charge types for the payments history page

- VAT inaccuracry assessment pen charge 
- VAT Misdeclaration penalty pre 2009 charge
- VAT MP Repeated Pre 2009 charge 
- VAT Inaccuracy Return Replaced charge
- VAT Wrongdoing penalty

### DIFF
--- a/test/views/templates/PaymentsHistoryChargeHelperSpec.scala
+++ b/test/views/templates/PaymentsHistoryChargeHelperSpec.scala
@@ -30,6 +30,13 @@ class PaymentsHistoryChargeHelperSpec extends UnitSpec {
           result shouldBe Some(paymentHistoryChargeType)
         }
       }
+
+      "return a Vat Inaccuracy Assessments Pen charge type" in {
+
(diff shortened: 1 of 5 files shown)


test/views/templates/PaymentsHistoryChargeHelperSpec.scala
```diff
@@ -30,6 +30,45 @@ class PaymentsHistoryChargeHelperSpec extends UnitSpec {
           result shouldBe Some(paymentHistoryChargeType)
         }
       }
+
+      "return a Vat Inaccuracy Assessments Pen charge type" in {
```
TheDeadlyPianist: Are these not tested by the above test?
Is it mandatory that we use hard coded strings, because the ChargeTypes were not tested with hard coded strings, they were done similarly to above (which is why I started automating the test).

TomDudley92: Completely agree. Forgot to take them out in the rebase

PR #363, opened 2019-01-24
BTAT-3938

lakshmipochampally:
Added the below chargeType = "VAT PA Default Interest"
mainType = "VAT PA Default Interest"
Charge name for heading = "Protective assessment default interest"
Label text = "interest charged on the protective assessment"


test/views/templates/PaymentsHistoryChargeHelperSpec.scala
```diff
@@ -31,6 +31,14 @@ class PaymentsHistoryChargeHelperSpec extends UnitSpec {
         }
       }
     }
+    "return a VAT PA default interest charge type" in {
```
TheDeadlyPianist: This is automatically tested by the test above, and can be removed.

PR #374, opened 2019-03-08
BTAT-5474 Added unallocated payment to payment history

TomDudley92:


app/models/viewModels/PaymentsHistoryModel.scala
```diff
@@ -30,50 +31,60 @@ case class PaymentsHistoryModel(chargeType: ChargeType,
...
-        list => if (list.nonEmpty) {
-          list.map(item => item.get[BigDecimal]("amount") ->
-            getOptionDate(item, FinancialTransactionsConstants.clearingDate))
-        } else {
-          throw new IllegalStateException("The items list was found but the list was empty")
+          val transactions: Seq[Option[PaymentsHistoryModel]] = getSubItemsForTransaction(transaction) map {
+            subItem => generatePaymentModel(chargeType, subItem, transaction)
+          }
+
+          transactions.flatten
```
sraspin: Could we filter out the transactions without a clearingDate please: `transactions.flatten.filter(_.clearingDate.isDefined)`

PR #381, opened 2019-03-15
BTAT-5476 Fixed {0} issue for charge descriptions that require a date

lewishay:
Some of the charge types require a date period (from - to) for their charge type descriptions. Currently if we don't receive period dates from the API, these descriptions render with a literal {0}. This fix logs the error and hides the description in those scenarios.

### DIFF
--- a/test/views/templates/payments/WhatYouOweChargeHelperSpec.scala
+++ b/test/views/templates/payments/WhatYouOweChargeHelperSpec.scala
@@ -42,35 +42,35 @@ class WhatYouOweChargeHelperSpec extends ViewBaseSpec {
 
   "WhatYouOweChargeHelper .description" when {
 
-    "payment has a to and from date" should {
+    "the charge has a to and from date" should {
 
       val helper = new WhatYouOweChargeHelper(paymentModel(ReturnDebitCharge), Some(false), messages)
 
-      "return the description of the payment" in {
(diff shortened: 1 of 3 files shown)


test/views/templates/payments/WhatYouOweChargeHelperSpec.scala
```diff
@@ -44,6 +51,16 @@ class WhatYouOweChargeHelperSpec extends ViewBaseSpec {
       }
     }
 
+    "payment should have a no to and from date to form the description, but doesn't" should {
```
davidmthornton: Can you reword this please so it's a bit easier to read.

test/views/templates/payments/WhatYouOweChargeHelperSpec.scala
```diff
@@ -44,6 +51,16 @@ class WhatYouOweChargeHelperSpec extends ViewBaseSpec {
...
+
+      val model = paymentModelNoPeriod(ReturnDebitCharge)
+      val helper = new WhatYouOweChargeHelper(model, Some(false), messages)
+
+      "omit the description of the payment" in {
+        helper.description shouldBe None
+      }
+    }
+
     "payment has a no to and from date" should {
```
davidmthornton: I know you didn't update this, but could you correct it please so it makes sense?

PR #384, opened 2019-03-25
[BTAT-5633] Controller gets mandation status, and changes front-end accordingly.

TheDeadlyPianist:


app/config/ConfigKeys.scala
```diff
@@ -42,6 +42,7 @@ object ConfigKeys {
   val useDirectDebitDummyPageFeature: String = "features.useDirectDebitDummyPage.enabled"
   val useLanguageSelectorFeature: String = "features.useLanguageSelector.enabled"
   val submitReturnFeatures: String = "features.submitReturnFeatures.enabled"
+  val checkMandationStatus: String = "features.checkMandationStatus.enabled"
```
TomDudley92: Please can you delete this as this feature switch already exists. It's the submitReturnFeatures one above.

app/controllers/VatDetailsController.scala
```diff
@@ -54,33 +56,36 @@ class VatDetailsController @Inject()(val messagesApi: MessagesApi,
       for {
         customerInfo <- accountDetailsCall
         nextReturn <- returnObligationsCall
-        nextPayment <- if(retrieveHybridStatus(customerInfo)) Future.successful(Right(None)) else paymentObligationsCall
+        nextPayment <- if (retrieveHybridStatus(customerInfo)) Future.successful(Right(None)) else paymentObligationsCall
+        mandationStatus <- mandationStatusService.getMandationStatus(user.vrn)
```
TomDudley92: Doesn't this need to be wrapped in the feature switch?

app/views/templates/nextReturnSection.scala.html
```diff
@@ -38,7 +38,11 @@ <h2 class="heading-medium" style="margin-top: 0px">@messages("returnObligation.n
 
   <p>
     <a id="vat-returns-link" href="@{appConfig.vatReturnDeadlinesUrl}" data-journey-click="return:view-deadlines:vat-details">
-      @messages("returnObligation.viewReturns")
+        @if(isNonMTDfB) {
```
TomDudley92: Shouldn't this be wrapped in a feature switch?

conf/messages
```diff
@@ -54,6 +54,7 @@ vatDetails.checkWhatYouOwe = Check what you owe and make a payment
 returnObligation.noReturn = No returns due right now
 returnObligation.nextReturn = Next return due
 returnObligation.viewReturns = View return deadlines
+returnObligation.submit = Submit return
```
TomDudley92: Do we need the Welsh Translation for this? Or will this be covered in another story?

app/views/templates/nextReturnSection.scala.html
TheDeadlyPianist: The `isNonMTDfB` is the switch. Is this not how it's done?

TheDeadlyPianist: If the feature switch is set to false, then this value will always be false.

app/controllers/VatDetailsController.scala
```diff
@@ -57,7 +57,8 @@ class VatDetailsController @Inject()(val messagesApi: MessagesApi,
         customerInfo <- accountDetailsCall
         nextReturn <- returnObligationsCall
         nextPayment <- if (retrieveHybridStatus(customerInfo)) Future.successful(Right(None)) else paymentObligationsCall
-        mandationStatus <- mandationStatusService.getMandationStatus(user.vrn)
+        mandationStatus <- if(appConfig.features.submitReturnFeatures()) {
+          mandationStatusService.getMandationStatus(user.vrn) } else { Future.successful(Right(MandationStatus("Disabled"))) }
```
TomDudley92: This else hasn't been covered within the tests. Can a quick test be added to the Controller Spec just cover this please?

TheDeadlyPianist:
@miranda-hawkes Assigned you to it because Tom told me to. It's the last comment he made that needs reviewing, and the Welsh content when I get it.

app/views/templates/nextReturnSection.scala.html
miranda-hawkes: There's a scenario in the story that you've missed:
When we can't retrieve the mandation status at all (and feature switch is on), put the following text in the partial:
```
Sorry, there is a problem with the service. 
Try again later.
```
and don't show the 'Submit return' or 'Return deadlines' link. See the story and google doc for the full requirement

it/pages/VatDetailsPageSpec.scala
```diff
@@ -41,15 +45,56 @@ class VatDetailsPageSpec extends IntegrationBaseSpec {
 
     "the user is authenticated" should {
 
-      "return 200" in new Test {
+      "return 200 and 'View returns deadline'" in new Test {
```
miranda-hawkes: Can you state 'and submit return feature switch is on' or something similar for these tests, and add (just one) scenario for when switch is off to prove the existing logic still stands. I know you tested that in the controller spec but if you're doing these scenarios at integration level you may as well add one with the switch off too

test/controllers/VatDetailsControllerSpec.scala
```diff
@@ -180,12 +188,50 @@ class VatDetailsControllerSpec extends ControllerBaseSpec {
...
+            .returning(mandationStatusServiceResult).once()
         }
 
         val result: Future[Result] = target.details()(fakeRequest)
         status(result) shouldBe Status.OK
       }
     }
+
+    "the feature switch is turned off" should {
+      "return a VatDetailsViewModel as a MTDfB user" in new DetailsTest {
```
miranda-hawkes: Saying 'return a VatDetailsViewModel' etc doesn't really reflect what the tests are asserting

app/views/templates/nextReturnSection.scala.html
TheDeadlyPianist: Oh, yeah. I opened the sub-task up, and didn't have the over-arching ticket open. My bad.

PR #387, opened 2019-04-12
BTAT-5746 Added previous payments tab; refactor to payments history view/cont

lewishay:


lewishay:
Temporarily closed due to redesign/refactor

PR #389, opened 2019-04-15
BTAT-5758 Added config for previous payments; updated customer info model

lewishay:


test/connectors/httpParsers/CustomerInfoHttpParserSpec.scala
```diff
@@ -16,9 +16,10 @@
 
 package connectors.httpParsers
 
+import common.TestModels.customerInformation
 import connectors.httpParsers.CustomerInfoHttpParser.CustomerInfoReads
 import models.errors._
-import models.{Address, CustomerInformation}
+import models.Address
```
sandypadman: unused  imports

PR #393, opened 2019-04-30
BTAT-5797 Update feedback form handoff

Jxr227:
- Now uses contact-frontend as the feedback form provider
- The following configs should be merged first:
https://github.com/hmrc/app-config-qa/pull/3865
https://github.com/hmrc/app-config-development/pull/2073
https://github.com/hmrc/app-config-staging/pull/2593
https://github.com/hmrc/app-config-production/pull/3566


app/config/frontendAppConfig.scala
```diff
@@ -80,7 +80,9 @@ trait AppConfig extends ServicesConfig {
   val portalPaymentHistoryUrl: String => String
   val portalNonHybridPreviousPaymentsUrl: String => String
   def languageMap:Map[String,Lang]
-  val routeToSwitchLanguage :String => Call
+  val routeToSwitchLanguage: String => Call
+  val host: String
```
lewishay: Could `host` be taken out of the `AppConfig` trait and instead just be a private val in `FrontendAppConfig`? Look like it's only `feedbackUrl` that needs to be accessible outside of here.

PR #415, opened 2019-06-07
BTAT-6084 Refactor to CustomerInformation model and updated VatSubscription route

lewishay:
As part of this task I had to update the VatSubscriptionConnector route to retrieve the full information in the same way the ChoC services do. This means that before this PR goes in, the subtask BTAT-6107 on the acceptance tests will need to be merged to ensure the acceptance tests don't fail due to stub data differences.

I have also removed the old AccountDetails components (controller/view) as I would have had to refactor those as part of this task, but as far as I'm aware they are unused.


it/stubs/CustomerInfoStub.scala
```diff
@@ -41,92 +41,74 @@ object CustomerInfoStub extends WireMockMethods {
...
-      |      "postCode":"GV13 4BJ"
-      |    },
-      |    "contactDetails":{
-      |      "primaryPhoneNumber":"01632 960026",
-      |      "mobileNumber":"07700 900018",
-      |      "emailAddress":"bettylucknexttime@gmail.com"
-      |    }
-      |  }
-      |}""".stripMargin
+  val customerInfoHybridUser: JsValue = Json.obj(
```
davidmthornton: Is there a way you can take a copy of customerInfoJsonMax and just update isPartialMigration? Or have a function to generate this data as it's quite hard to distinguish the difference as it it?

it/stubs/CustomerInfoStub.scala
```diff
@@ -41,92 +41,74 @@ object CustomerInfoStub extends WireMockMethods {
...
-      |      "postCode":"GV13 4BJ"
-      |    },
-      |    "contactDetails":{
-      |      "primaryPhoneNumber":"01632 960026",
-      |      "mobileNumber":"07700 900018",
-      |      "emailAddress":"bettylucknexttime@gmail.com"
-      |    }
-      |  }
-      |}""".stripMargin
+  val customerInfoJsonMax: JsValue = Json.obj(
```
davidmthornton: Is it possible to get this from test/common/TestJson.scala instead of repeating here?

lewishay: It picks up the import but then errors when the tests run. I'm not sure if it fails because it's in a different test package. I tried using `_root_` in front of the import too but got the same issue. If you know how to do it then let me know.

davidmthornton: I've had a go too but no dice! It's fine as is 👍

davidmthornton:
This looks good to go once ATs are in and the view-vat-returns issue has been sorted.

PR #425, opened 2019-06-24
BTAT-6165 Create feature Switch and add opt out section on Vat summary

sandypadman:


conf/application.conf
```diff
@@ -126,6 +126,7 @@ features {
   allowDirectDebits.enabled = true
   staticDate.enabled = true
   vatCertificate.enabled = true
+  vatOptOutService.enabled=true
```
lewishay: Could you also make an `app-config-base` PR to add the new config there and blank out the `vat-opt-out-frontend` host, and an `app-config-production` PR to set this feature switch to false.

conf/messages
```diff
@@ -49,6 +49,9 @@ vatDetails.title = Your VAT account
 vatDetails.manageYourVat = Manage your VAT
 vatDetails.vatCertificate.view = View and print your VAT certificate.
 vatDetails.vatCertificate.link = View VAT certificate
+vatDetails.vatOptOut.view = You cannot opt out if your taxable turnover has been above £85,000 since 1 April 2019.
+vatDetails.vatOptOut.link = Opt out of Making Tax Digital for VAT
+vatDetails.vatOptOut.error = Sorry, there is a problem with the service. Try again later.
```
lewishay: Do we have Welsh content for this message? If not at the moment that's fine as we discussed not blocking stories anymore due to Welsh, but could you check if there is a story to do this work, and if not raise a new one?

test/views/templates/VatOptOutSectionTemplateSpec.scala
```diff
@@ -0,0 +1,51 @@
...
+
+    object Selectors {
+      val vatOptOutHeading = "#vat-optout > h2:nth-child(1) a"
+      val vatOptOutParagraph = "#vat-optout > p:nth-child(2)"
+    }
+
+    def view: HtmlFormat.Appendable = views.html.templates.vatOptOutSection(Some(false))
+    implicit def document: Document = Jsoup.parse(view.body)
+
+    "the vat opt out feature" should {
```
lewishay: This `should` clause seems unnecessary. Might be worth just changing the `when` above to a `should` and then removing this extra clause.

app/testOnly/views/featureSwitch.scala.html
```diff
@@ -34,6 +34,7 @@ <h1 class="heading-medium">Features</h1>
                 @views.html.templates.inputs.singleCheckbox(form(ConfigKeys.allowDirectDebitsFeature), "Direct debits allowed")
                 @views.html.templates.inputs.singleCheckbox(form(ConfigKeys.staticDateEnabledFeature), "Use static date (2018-05-01)")
                 @views.html.templates.inputs.singleCheckbox(form(ConfigKeys.vatCertificateFeature), "Show vat certificate")
+                @views.html.templates.inputs.singleCheckbox(form(ConfigKeys.vatOptOutServiceFeature), "Show vat opt-out service")
```
lewishay: Could you change this feature switch description to something like "Show vat opt-out section", just to be clear on what it is for others who see it.

test/views/vatDetails/VatDetailsViewSpec.scala
```diff
@@ -372,4 +404,26 @@ class VatDetailsViewSpec extends ViewBaseSpec {
       element(Selectors.header).attr("data-metrics") shouldBe "error:hidden-text:vat-overview-entity-name"
     }
   }
+
+  "Rendering the error message with error in Customer information API" should {
+
+    lazy val view = views.html.vatDetails.details(bothErrorDetailsModel)
+    lazy implicit val document: Document = Jsoup.parse(view.body)
+
+    "render Api graceful error message in Opt out section" in {
```
lewishay: Could you move this test to the clause above it which is regarding errors in all APIs. It constructs the view with the same view model you are using here so it should work the same.

test/views/vatDetails/VatDetailsViewSpec.scala
```diff
@@ -106,7 +116,16 @@ class VatDetailsViewSpec extends ViewBaseSpec {
     currentYear,
     paymentError = true,
     returnObligationError = true,
-    isNonMTDfBUser = None
+    isNonMTDfBUser = None,
+    customerInfoError = true
+  )
+  val customerNonMtdModel = VatDetailsViewModel(
```
lewishay: Instead of making a new model here could you use the `copy` method. For example if you use `detailsModel.copy()` and then just provide the one parameter that you are changing (here it would be `isNonMTDfBUser = Some(true)`)

Ideally the other models here would make use of copy as it's getting a bit bloated but don't worry about those, that could be a future refactor.

app/views/vatDetails/details.scala.html
```diff
@@ -17,6 +17,7 @@
 @import models.viewModels.VatDetailsViewModel
 @import views.html.templates.formatters.breadcrumbs._
 @import views.html.templates.formatters.links._
+@import common.SessionKeys
```
lewishay: This looks like it isn't being used

test/views/vatDetails/VatDetailsViewSpec.scala
lewishay: Still unresolved

PR #429, opened 2019-07-04
BTAT-6217 Updated payments sub item parser to read 'paymentAmount'

lewishay:
The unit and IT test data for FinancialData has also been slightly updated to be more accurate with regards to how the real data is returned.

**NOTE** - this change will break the acceptance tests until the stub data has been updated as part of BTAT-6214


davidmthornton:
This looks good to go in - but will wait for BTAT-6214 to go in.

PR #440, opened 2019-08-07
BTAT-6451 Added dynamic exit survey identifier based on user type

lewishay:


app/controllers/SignOutController.scala
```diff
@@ -18,23 +18,39 @@ package controllers
...
-    val redirectUrl: String = if (authorised) appConfig.signOutUrl else appConfig.unauthorisedSignOutUrl
-    Future.successful(Redirect(redirectUrl))
+    implicit val hc: HeaderCarrier =
+      HeaderCarrierConverter.fromHeadersAndSession(request.headers, Some(request.session))
+
+    if(authorised) {
+      enrolmentsAuthService.authorised.retrieve(Retrievals.affinityGroup) {
+        case Some(AffinityGroup.Agent) => Future.successful("VATCA")
+        case Some(AffinityGroup.Individual | AffinityGroup.Organisation) => Future.successful("VATC")
+        case _ => Future.successful("VATCU")
```
Jxr227: Do these need Future.successful()? Worried I'm missing something obvious so put me in my place if needs be. But if they're all using it can we just stop this by an asynchronous action?

lewishay: Resolved in person

PR #455, opened 2019-09-18
BTAT-6817 - What You Owe page changed to use definition lists

kylemchownhmrc:


app/views/templates/payments/whatYouOweChargeRow.scala.html
```diff
@@ -24,42 +24,42 @@
...
+            </div>
+        </dt>
+        <dd class="cya-answer what-you-owe-amount">
+            <span data-amount="@(payment.amount)">@displayMoney(payment.amount)</span>
+        </dd>
+        <dd class="cya-change what-you-owe-links">
+            <div>
+                <a data-journey-click="payment:make-payment:open-payments" href="@payment.makePaymentRedirect">
+                    <span>@chargeDetails.payLinkText</span>
+                    <span class="visuallyhidden">
```
lewishay: One of the issues DAC raised is that the visually hidden screen reader content for the "Pay now" link was too verbose and that we only needed the amount. Could you remove the two unnecessary lines from this `visuallyhidden` span please.

Could you also remove the val `overdueContext` from `WhatYouOweChargeHelper`.

PR #456, opened 2019-09-20
BTAT-6748 Added accessibility statement and footer link

lewishay:


conf/messages
```diff
@@ -308,4 +308,67 @@ returnPeriod.MM = Every month
 returnPeriod.MA = April, July, October and January
 returnPeriod.MB = May, August, November and February
 returnPeriod.MC = March, June, September and December
-returnPeriod.nonStandard = Non-standard
\ No newline at end of file
+returnPeriod.nonStandard = Non-standard
```
paulfirmstone: I think we need to add content for the following:
On the ‘What you owe’ page, some headings were duplicated and weren’t descriptive enough so they were challenging to understand for screen reader users - AA
This doesn’t meet WCAG ‘2.4.6 Headings and Labels’

PR #474, opened 2019-11-22
BTAT-7176

kylemchownhmrc:


app/models/viewModels/VatDetailsViewModel.scala
```diff
@@ -27,6 +27,6 @@ case class VatDetailsViewModel(paymentsData: Option[String],
                                paymentOverdue: Boolean = false,
                                paymentError: Boolean = false,
                                isHybridUser: Boolean = false,
-                               isNonMTDfBUser: Option[Boolean] = Some(false),
+                               isNonMTDfBOrNonDigitalUser: Option[Boolean] = Some(false),
```
lewishay: The details view is not compiling because it's still referencing the old name of this variable.

PR #487, opened 2019-12-24
Btat 7410

JayKL:


app/config/ServiceErrorHandler.scala
```diff
@@ -14,16 +14,38 @@
  * limitations under the License.
  */
 
+/*
```
miranda-hawkes: Please remove duplicate header

PR #488, opened 2019-12-27
BTAT-7419: Removed year from returns history hand off

davidmthornton:
### Needs to be merged with BTAT-7420 (view-vat-returns-frontend)


davidmthornton:
**### Needs to be merged with BTAT-7420 (view-vat-returns-frontend)**

PR #494, opened 2020-02-06
[BTAT-7522] - added dereg partial to vat-overview page

sraspin:


sraspin:
https://github.com/hmrc/app-config-base/pull/3088

PR #496, opened 2020-03-03
BTAT-7584 - Lanuage toggle moved above back link

kylemchownhmrc:


public/stylesheets/palette.css
```diff
@@ -273,3 +273,7 @@ input.form-control-error[type="password"],
 .notice-banner {
   background-color: #1d70b8;
 }
+
+.text--right {
```
lewishay: Can you move this to our general-purpose VAT CSS file instead. `palette.css` is for changes regarding the colour palette.

app/views/govuk_wrapper.scala.html
```diff
@@ -127,11 +127,11 @@
 
 @mainContentHeader = {
     @if(appConfig.features.useLanguageSelector()){
-    <div class="float--right">@views.html.language_selection(
-        appConfig.languageMap,
-        appConfig.routeToSwitchLanguage,
-        customClass = None
-    )</div>
+        @views.html.language_selection(
```
lewishay: `manage-vat-subscription-frontend` has the language toggle above the BTA nav links, but here on this service the BTA links are above the toggle. I imagine we want to standardise this. To me it makes more sense to have the language toggle above the links but it would be worth asking Barry what he thinks.

kylemchownhmrc: When I checked with Barry he looked at how it was done on some pages we don't own and it has the toggle below the BTA links so he said we should do the same. This does mean we'd need to edit the other service at some point though

PR #497, opened 2020-03-12
[BTAT-7618] - added feature switch for DDCollection flag

sraspin:


sraspin:
https://github.com/hmrc/app-config-base/pull/3213
https://github.com/hmrc/app-config-production/pull/6023

PR #501, opened 2020-03-19
[BTAT-7665] - changed to use vat-subscription-dynamic-stub instead of…

sraspin:
… vat-vc-api-stub

### DIFF
--- a/conf/application.conf
+++ b/conf/application.conf
@@ -95,7 +95,7 @@ microservice {
     }
     vat-subscription {
       host = localhost
-      port = 9156
+      port = 9567
     }
     selfLookup {
       host = localhost
--- a/conf/application.conf
+++ b/conf/application.conf
@@ -75,7 +75,7 @@ microservice {
     }
     vat-api {
       host = localhost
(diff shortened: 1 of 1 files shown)


sraspin:
https://github.com/hmrc/app-config-base/pull/3256
https://github.com/hmrc/app-config-staging/pull/4548

conf/application.conf
```diff
@@ -95,7 +95,7 @@ microservice {
     }
     vat-subscription {
       host = localhost
-      port = 9154
+      port = 9156
```
lewishay: This needs to be `9567`

conf/application.conf
```diff
@@ -75,7 +75,7 @@ microservice {
     }
     vat-api {
       host = localhost
-      port = 9154
+      port = 9156
```
lewishay: I see what you mean about this. It's not even being used as its the "off" case for an old feature switch that has been turned on for ages. I think for now this is fine and we can remove this `vat-api` config as part of that feature switch removal down the line?

PR #513, opened 2020-05-12
[DM] [BTAT-7793] Updated mandation statuses behind fs + content

TheDeadlyPianist:
The following MUST be merged first:
- [x] [App-Config Base](https://github.com/hmrc/app-config-base/pull/3563)
- [x] [App-Config Production](https://github.com/hmrc/app-config-production/pull/6886)


app/config/ConfigKeys.scala
```diff
@@ -43,6 +43,7 @@ object ConfigKeys {
   val mtdSignUpFeature: String = "features.mtdSignUp.enabled"
   val ddCollectionInProgressFeature: String = "features.ddCollectionInProgress.enabled"
   val displayCovidMessageFeature: String = "features.displayCovid.enabled"
+  val mergedMTDfBStatuses: String = "features.mergedMTDfBStatuses.enabled"
```
lewishay, line 46: There is no mention of a feature switch in the story, and following the discussion in our team channel I don't think one is necessary.

app/controllers/VatDetailsController.scala
```diff
@@ -163,19 +165,27 @@ class VatDetailsController @Inject()(val messagesApi: MessagesApi,
...
     )
   }
 
-  private def retrieveIsNonMTDfBOrNonDigital(mandationStatus: HttpGetResult[MandationStatus]): Option[Boolean] = {
-    mandationStatus.fold(
-      _ => None,
-      result => Some(result.mandationStatus == nonMTDfB || result.mandationStatus == nonDigital)
-    )
-  }
+  private def retrieveIsExempt(mandationStatus: HttpGetResult[MandationStatus]): Option[Boolean] =
```
lewishay, line 181: Could we simplify these three functions into one? As far as I'm aware, at the moment in this service we treat these three statuses exactly the same (correct me if I'm wrong). Just having one function here and one boolean in the view model would simplify all this

conf/messages
```diff
@@ -58,6 +58,7 @@ vatDetails.vatOptOut.info = You cannot opt out if your taxable turnover has been
 vatDetails.vatOptOut.pending = Your request to opt out is being processed.
 vatDetails.mtdSignup.link = Sign up for Making Tax Digital for VAT
 vatDetails.mtdSignup.info = If your taxable turnover exceeds the VAT threshold, you must sign up to Making Tax Digital for VAT.
+vatDetails.mtdSignup.info.newStatuses = You must sign up to Making Tax Digital for VAT if you''re not exempt from VAT, and your taxable turnover exceeds the £85,000 threshold.
```
lewishay, line 61: As mentioned above, as the feature switch is not needed you can just update the original message with this content. The welsh is okay to be left as is and will be played in a separate story.

conf/application.conf
```diff
@@ -136,6 +136,7 @@ features {
   mtdSignUp.enabled = true
   ddCollectionInProgress.enabled = true
   displayCovid.enabled = true
+  mergedMTDfBStatuses.enabled = true
```
lewishay, line 139: This needs stripping out too

conf/messages
```diff
@@ -57,7 +57,7 @@ vatDetails.vatOptOut.link = Opt out of Making Tax Digital for VAT
 vatDetails.vatOptOut.info = You cannot opt out if your taxable turnover has been above £85,000 since 1 April 2019.
 vatDetails.vatOptOut.pending = Your request to opt out is being processed.
 vatDetails.mtdSignup.link = Sign up for Making Tax Digital for VAT
-vatDetails.mtdSignup.info = If your taxable turnover exceeds the VAT threshold, you must sign up to Making Tax Digital for VAT.
+vatDetails.mtdSignup.info = You must sign up to Making Tax Digital for VAT if you''re not exempt from VAT, and your taxable turnover exceeds the £85,000 threshold.
```
lewishay, line 60: Can you use a smart apostrophe here instead of two straight ones. It renders slightly differently and you don't have to escape it. We should have replaced all of those a while ago (looks like a few are still lingering around unfortunately).

app/controllers/VatDetailsController.scala
```diff
@@ -163,17 +160,16 @@ class VatDetailsController @Inject()(val messagesApi: MessagesApi,
     }
   }
 
-  private def retrieveIsNonMTDfB(mandationStatus: HttpGetResult[MandationStatus]): Option[Boolean] = {
+  private[controllers] def retrieveIsOfStatus(mandationStatus: HttpGetResult[MandationStatus], expectedType: String*): Option[Boolean] = {
```
lewishay, line 163: Can you change `expectedType` to just be a `Seq[String]` rather than using varargs, unless I'm missing something.

TheDeadlyPianist: Was for a reason when those other functions existed. Doesn't make a difference now. Will remove.

PR #527, opened 2020-07-10
BTAT-7966-1 Made corrections to email unverified nudge work

lewishay:
Please ensure the following are merged:

- https://github.com/hmrc/app-config-base/pull/3806
- https://github.com/hmrc/app-config-production/pull/7573

### DIFF
--- a/app/controllers/VatDetailsController.scala
+++ b/app/controllers/VatDetailsController.scala
@@ -30,7 +30,6 @@ import models._
 import models.obligations.{Obligation, VatReturnObligation, VatReturnObligations}
 import models.payments.{Payment, Payments}
 import models.viewModels.VatDetailsViewModel
-import org.slf4j
 import play.api.Logger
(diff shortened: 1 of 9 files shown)


app/controllers/VatDetailsController.scala
```diff
@@ -204,16 +199,16 @@ class VatDetailsController @Inject()(val enrolmentsAuthService: EnrolmentsAuthSe
...
-            case (Some(_), Some(verified)) => verified
-            case (Some(_), None) => false
-            case _ => true
-          }
-        case _ => true
-      })
+    accountDetails match {
+      case Right(details) =>
+        details.emailAddress match {
+          case _ if details.hasPendingPpobChanges => true
```
sraspin, line 205: This will prevent users with any ppob inflight changes from seeing the email nudge - is that what we want or do we want to do this only for inflight email changes?

PR #529, opened 2020-07-23
[BTAT-8036] Added host to URLs in frontendAppConfig

JoNB7777:


app/config/frontendAppConfig.scala
```diff
@@ -173,8 +173,8 @@ class FrontendAppConfig @Inject()(val runModeConfiguration: Configuration, sc: S
   override lazy val directDebitRedirectUrl: String = sc.getString(Keys.directDebitRedirectUrl)
 
   private lazy val paymentsReturnBase: String = sc.getString(Keys.paymentsReturnBase)
-  override lazy val paymentsReturnUrl: String = paymentsReturnBase + sc.getString(Keys.paymentsReturnUrl)
-  override lazy val paymentsBackUrl: String = paymentsReturnBase + sc.getString(Keys.paymentsBackUrl)
+  override lazy val paymentsReturnUrl: String = host + paymentsReturnBase + sc.getString(Keys.paymentsReturnUrl)
```
lewishay, line 176: These config values are used for the journey of making a payment. The config values you need to change for this story are `directDebitReturnUrl` and `directDebitBackUrl`. You can see these two values being used in the action of `DirectDebitController`.

The change should also be to replace the "base" in those values with `host`, rather than just prefix `host`. This is because _locally_ the DD base is actually set correctly (line 229 of `application.conf`), so doing it this way you would end up with a duplicate host locally.

PR #533, opened 2020-08-05
BTAT-8131 content change

SheilaTran:


Tasmine10ten:
Overall change looks good & works manually. Just need to amend a couple of unit tests

PR #540, opened 2020-08-25
[DM] [BTAT-8205] Users can now check payments from last 24 months

TheDeadlyPianist:


JBlacklawsTen10 (approved):
Passes locally and modified ATs as well as unmodified PTs

PR #542, opened 2020-09-08
Btat 8033 Scala and sbt upgrade

JoNB7777:
Review and merge together with https://github.com/hmrc/service-manager-config/pull/2565#issue-482135449 and https://github.com/hmrc/build-jobs/pull/6491#issue-482135204


project/build.properties
```diff
@@ -1 +1 @@
-sbt.version=0.13.18
+sbt.version=1.3.4
```
paulfirmstone, line 1: There is now a later version which we might aswell upgrade to. I believe I used 1.3.13 for VACLUF.

build.sbt
```diff
@@ -95,7 +97,7 @@ lazy val microservice: Project = Project(appName, file("."))
   .settings(
```
paulfirmstone, line 97: Can youy add .disablePlugins(JUnitXmlReportPlugin) to avoid the unstable build issue.

paulfirmstone (changes requested):
Putting aside the warning about *. in the views there are still a handful that can be addressed in the app. There also a few warnings in the tests that can be looked at too.

PR #544, opened 2020-09-10
Added 'Payment recovery' charge type

daparkes:


JBlacklawsTen10 (approved):
New charge type appears correctly, passes local tests and all ATs and PTs

PR #561, opened 2021-01-15
[BTAT-8832] - gets the customers HybridToFull date instead of migratedToETMP date if it exists

sraspin:


SheilaTran (approved):

SheilaTran (approved):

lewishay (approved):
Approved awaiting ATs

PR #562, opened 2021-01-21
BTAT-8853 Implemented logic to block future insolvent users from payment pages

lewishay:
As part of this, I've replaced the old Hybrid predicate with a FinancialPredicate which checks both hybrid status and future insolvency to reduce the number of API calls, as there was no easy way to just add on the new work without adding another call. The two traits work well together as they are both checked in the same actions.

I also made use of another session key, purely for the happy flow users to prevent making repeated calls to check hybrid & insolvency date. I didn't go into the "false" case logic because it would have been quite complex depending on whether the user was hybrid or insolvent.


SheilaTran (approved):

SheilaTran:
Waiting for BTAT-8854 (AT)

PR #563, opened 2021-01-28
[BTAT-8878] Repayment Supplement Recovery charge type

daparkes:


test/common/MessageLookup.scala
```diff
@@ -41,6 +41,7 @@ object MessageLookup {
       val datePeriod = if(useLongDateFormat) datePeriodLong else datePeriodShort
 
       chargeType match {
+        case RepaymentSupplementRecovery.name => ("VAT repayment supplement recovery", "to recover the amount you were overpaid")
```
paulfirmstone, line 44: This is causing a test to fail as its still checking for the VAT prefix.

paulfirmstone (changes requested):

PR #568, opened 2021-04-09
[BTAT-9134] - Prevent Unallocated POA appearing as a refund

william-hatzar:


app/models/viewModels/PaymentsHistoryModel.scala
```diff
@@ -46,7 +48,8 @@ object PaymentsHistoryModel {
 
           val chargeType: ChargeType =  ChargeType.apply((transaction \ FinancialTransactionsConstants.chargeType).as[String])
 
-          val transactions: Seq[Option[PaymentsHistoryModel]] = getSubItemsForTransaction(transaction) map {
+          val transactions: Seq[Option[PaymentsHistoryModel]] = getSubItemsForTransaction(transaction)
+            .filterNot(_.clearingReason.contains(allocatedCharge)) map {
```
sraspin, line 52: I've been thinking it might be worth changing this to _.clearingReason.map(_.toLowerCase).contains... and change allocatedCharge to "allocated to charge". Just to be on the safe side

PR #574, opened 2021-05-04
[BTAT-9170] - Remove api call from open payments controller

william-hatzar:


test/controllers/OpenPaymentsControllerSpec.scala
```diff
@@ -295,10 +295,6 @@ class OpenPaymentsControllerSpec extends ControllerBaseSpec {
         "return the payments view" in new Test {
           override def setupMocks(): Unit = {
             super.setupMocks()
-            (mockPaymentsService.getDirectDebitStatus(_: String)(_: HeaderCarrier, _: ExecutionContext))
```
paulfirmstone, line 298: This mock is setup in a number of tests and I don't believe it is needed. Can you double check please?
Also DirectDebitStatusError is an unused import in test test spec.

paulfirmstone (changes requested):

PR #576, opened 2021-05-06
[BTAT-9212] Added dummy views for DD interrupt

daparkes:


JoNB7777 (changes requested):
The changes you made look all good, but the parameter user in the 'NotFound' error view doesn't seem to be required anymore so could maybe still remove that?

PR #589, opened 2021-06-28
BTAT_9344-Refactor Direct Debit Interrupt Logic

william-hatzar:


test/controllers/OpenPaymentsControllerSpec.scala
```diff
@@ -30,6 +30,7 @@ import org.jsoup.Jsoup
 import org.jsoup.nodes.Document
 import play.api.http.Status
 import play.api.mvc.{Request, Result}
+import play.api.test.FakeRequest
```
JoNB7777, line 33: I think this import is unused as all tests in this Spec that are using a FakeRequest are using ones defined in ControllerBaseSpec

test/controllers/OpenPaymentsControllerSpec.scala
```diff
@@ -49,6 +50,8 @@ class OpenPaymentsControllerSpec extends ControllerBaseSpec {
     val serviceInfoServiceResult: Future[Html] = Future.successful(Html(""))
     val accountDetailsResponse: HttpGetResult[CustomerInformation] = Right(customerInformationMax)
     val mockServiceInfoService: ServiceInfoService = mock[ServiceInfoService]
+    val accountDetailsServiceResult: Future[HttpGetResult[CustomerInformation]] =
```
JoNB7777, line 53: I believe this is not actually used anywhere in this spec either

JoNB7777 (changes requested):

JoNB7777 (approved):

PR #590, opened 2021-07-02
Btat 9386 Moved Back Links out of MainContent

JoNB7777:


paulfirmstone (changes requested):
The vat certificate page also has a back link for Agents so will also need to be changed.

Although the title of the JIRA story refers to back links there is also mention of breadcrumbs so they should also be moved outside of the main content. We have confirmed this with plat-ui.

PR #594, opened 2021-07-13
[BTAT-9429] Fixed how user research banner is being displayed

JoNB7777:


app/views/MainTemplate.scala.html
```diff
@@ -108,9 +108,9 @@
     serviceName = Some(dynamicNavTitle),
     signOutUrl = if(showLogout) Some(routes.SignOutController.signOut(feedbackOnSignOut).url) else None,
     phaseBanner = Some(standardBetaBanner(url = appConfig.feedbackUrl(request.uri))),
+    userResearchBanner = if(appConfig.features.userResearchBanner()) Some(UserResearchBanner(url = "https://signup.take-part-in-research.service.gov.uk/?utm_campaign=VATviewchange&utm_source=Other&utm_medium=other&t=HMRC&id=34")) else None,
```
daparkes, line 111: The banner is still displaying a little wonkily. Sheila had the same issue and the solution we found was to change the custom CSS we have for the banner (`.hmrc-user-research-banner`) from `margin-top: -10px` to `margin-bottom: -5px`. If you do that it should move down and stop blocking the black bar at the top

daparkes (changes requested):

PR #599, opened 2021-07-15
[BTAT-9457] Removed submitReturn feature switch

JoNB7777:


test/controllers/VatDetailsControllerSpec.scala
```diff
@@ -236,25 +233,13 @@ class VatDetailsControllerSpec extends ControllerBaseSpec {
...
-        override val accountDetailsServiceResult: Future[HttpGetResult[CustomerInformation]] =
-          Future.successful(Right(customerInformationNonMTDfB))
+    "return a VatDetailsViewModel as a non MTDfB user" in new DetailsTest {
+      override val accountDetailsServiceResult: Future[HttpGetResult[CustomerInformation]] =
+        Future.successful(Right(customerInformationNonMTDfB))
 
-        mockAppConfig.features.submitReturnFeatures(true)
-        lazy val result: Future[Result] = target().details()(fakeRequest)
-        status(result) shouldBe OK
-        await(bodyOf(result)).contains(messages("returnObligation.submit")) shouldBe true
```
william-hatzar, line 256: I think you should keep this test as the feature switch is on so it will be testing code that is still there. All you need to remove is mockAppConfig part where it sets it to true

william-hatzar (approved):

PR #600, opened 2021-07-15
[BTAT-9458] Removed agentAccess feature switch

JoNB7777:


test/controllers/VatCertificateControllerSpec.scala
```diff
@@ -61,8 +61,6 @@ class VatCertificateControllerSpec extends ControllerBaseSpec {
         .stubs(*, *, *)
```
daparkes, line 61: There's a variable on line 46 in here (github won't let me comment that line) that you could get rid of. I think it was only used for the feature switch stuff

daparkes (changes requested):

daparkes (approved):

PR #602, opened 2021-07-15
[BTAT-9455] Remove useDirectDebitDummyPage feature switch

daparkes:
I think all the direct debit components in the service were unused after this feature switch was taken out (they were related to a handoff we used to have in the service and got rid of). So I've removed them as well as the feature switch


lewishay (changes requested):
Can you please also remove:
- `DirectDebitRedirectUrlHttpParser` & associated tests
- `postSetupDirectDebit...` elements from `MetricsService`
- `DirectDebitSetupError` from the service error models

It looks to me like these were all used as part of the now-deleted components

lewishay (approved):

PR #608, opened 2021-07-27
[BTAT-9474]-Remove Play UI from VAT Summary

william-hatzar:


daparkes (changes requested):
I think where we landed on this was we would unit test the component we brought in from play-ui. Can you copy in some tests from the bottom of here: https://github.com/hmrc/submit-vat-return-frontend/pull/170/files (I just copied these from play-ui and it looks like yours would be the same apart from the file name)

PR #610, opened 2021-08-12
BTAT-9581 CSS fix for BTA links

lewishay:
The javascript is now unnecessary as BTA are on the new assets. The CSS I used is the same styling that the `ul`/`li`/`a` elements in the old BTA partial were using, with the exception of the font-size tweak for small screens I've added to prevent messiness and preserve the single horizontal row.


lewishay:
Closed due to this being addressed on BTA's side

PR #616, opened 2021-09-02
[BTAT-9601] - Added in service url to header link so it redirects back to start

william-hatzar:


JoNB7777 (approved):
Waiting for pipeline to be clear before this can be merged

PR #621, opened 2021-09-30
BTAT-9679 Updating no DD interrupt content

paulfirmstone:


paulfirmstone:
Please merge with: https://github.com/hmrc/view-vat-acceptance-tests/pull/288

app/views/interrupt/DDInterruptNoDD.scala.html
```diff
@@ -15,7 +15,7 @@
  *@
 
 @this(mainTemplate: MainTemplate,
-        govukDetails: GovukDetails,
+        govukInsetText : GovukInsetText,
```
lewishay, line 18: This isn't something you have changed but I've just seen that the `Back` link has no Welsh translation. I believe this is because it does not have its `content` parameter set, so it defaults  to "Back" regardless of language. Could you correct this while you're here?

lewishay (changes requested):
I have a question about the inset text content as well as mentioned in Slack

PR #623, opened 2021-10-06
Btat 9709 Added agent handoff message to Payment history page

JoNB7777:


test/views/payments/PaymentHistoryViewSpec.scala
```diff
@@ -224,110 +226,129 @@ class PaymentHistoryViewSpec extends ViewBaseSpec {
...
+      "not display the insolvency banner" in {
+        elementExtinct(Selectors.insolvencyBanner)
+      }
     }
+  }
 
-    "not have the 'Previous payments' tab content, despite the previousPaymentsTab boolean being set to true" in {
-      elementExtinct(Selectors.prevPaymentsParagraph)
-    }
+  "the client's hybridToFullMigrationDate is not within the last three years" should {
```
lewishay, line 343: Should be "is within" rather than "is not within"

lewishay (changes requested):

PR #631, opened 2021-10-12
Btat 9729 Refactored remaining unit test specs

JoNB7777:


test/connectors/ServiceInfoPartialConnectorSpec.scala
```diff
@@ -32,45 +32,78 @@ import scala.concurrent.{ExecutionContext, Future}
...
+  val httpClient: HttpClient = mock[HttpClient]
 
   "ServiceInfoPartialConnector" should {
-    "generate the correct url" in new Test {
+
+    "generate the correct url" in {
+
+      lazy val connector: ServiceInfoPartialConnector = {
+
+        (httpClient.GET[HtmlPartial](_: String, _: Seq[(String, String)], _: Seq[(String, String)])
```
lewishay, line 45: Could you put this mocking and connector creation in a little function in this file? Just conscious it's being repeated in every test here with the only change being the value that is returned.

test/services/AccountDetailsServiceSpec.scala
```diff
@@ -32,28 +32,24 @@ import scala.concurrent.ExecutionContext.Implicits.global
...
+    new AccountDetailsService(mockVatSubscriptionConnector)
   }
 
   "Calling .getAccountDetails" should {
 
-    "retrieve the customer details" in new Test {
-      val details: HttpGetResult[CustomerInformation] = await(target.getAccountDetails("123456789"))
+    "retrieve the customer details" in {
+      val details: HttpGetResult[CustomerInformation] = {
+        (mockVatSubscriptionConnector.getCustomerInfo(_: String)(_: HeaderCarrier, _: ExecutionContext))
```
lewishay, line 48: Could you put this mocking in a little function in this file? Just conscious it's being repeated in every test here with the only change being the value that is returned (I don't think the VRN matters).

lewishay (approved):

PR #632, opened 2021-10-12
BTAT-9728 Refactor of VatDetailsServiceSpec

paulfirmstone:


lewishay (approved):

test/services/VatDetailsServiceSpec.scala
```diff
@@ -31,8 +31,7 @@ import play.api.test.Helpers.{await, defaultAwaitTimeout}
 import scala.concurrent.{ExecutionContext, Future}
 
 class VatDetailsServiceSpec extends ControllerBaseSpec {
-
-  private trait Test {
+
```
lewishay, line 34: As mentioned on Slack, if you could remove most of the payment models from this spec to reduce bloat as only one is needed to prove the test

lewishay (approved):

PR #638, opened 2021-11-09
[BTAT-9845] Added new error view for insolvent users

JoNB7777:


app/views/errors/UserInsolventError.scala.html
```diff
@@ -0,0 +1,30 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *@
+
+@this(mainTemplate: MainTemplate, govukButton: GovukButton)
+
+@()(implicit request: Request[_], messages: Messages, appConfig: config.AppConfig)
+
+@mainTemplate(title = messages("insolventError.title"), appConfig = appConfig) {
+    <h1 class="govuk-heading-xl" id="insolvent-without-access-heading">@messages("insolventError.title")</h1>
```
lewishay, line 22: This is just an assumption based on the screenshot, but to me it looks like the design has a smaller heading and a standard paragraph size, rather than the larger ones you've used.

Might be worth asking Bobby to confirm this or provide a link to the page on the prototype if we can't agree based on the screenshot alone.

lewishay (changes requested):

PR #641, opened 2021-11-18
[BTAT-9876] Improve test coverage

daparkes:


test/controllers/VatDetailsControllerSpec.scala
```diff
@@ -263,20 +263,36 @@ class VatDetailsControllerSpec extends ControllerBaseSpec {
...
   }
 
-  "Calling .detailsRedirectToEmailVerification" should {
+  "Calling .detailsRedirectToEmailVerification" when {
 
-    "redirect to email verification" when {
+    "all relevant information is returned from vat-subscription" when {
 
-      "all relevant information is returned from vat-subscription" in {
+      "there is no pending PPOB change" should {
```
lewishay, line 270: Can you add one extra test to each of these scenarios (no pending & pending) to check the `inFlightContactKey` is being added or not added to session, as that seems to be the difference between these two scenarios.

Bonus comment - can you remove the unused import in this file

app/controllers/VatDetailsController.scala
```diff
@@ -265,6 +262,7 @@ class VatDetailsController @Inject()(vatDetailsService: VatDetailsService,
                                        paymentObligations: ServiceResponse[Option[Payments]])
                                       (implicit hc: HeaderCarrier): Unit = {
 
+    //$COVERAGE-OFF$
```
lewishay, line 265: Why were these added?

app/models/payments/ChargeType.scala
```diff
@@ -276,7 +276,9 @@ object ChargeType {
       true
     } catch {
       case ex: IllegalArgumentException =>
+        //$COVERAGE-OFF$
```
lewishay, line 279: I wonder if this logger message was flagged up due to the fact we are not using our shared application logger here, but instead creating a new logger inside of the `ChargeType` object? Can you make the object extend `LoggerUtil` and see if the coverage complaint goes away? Even if not, it's good for us to reuse our existing logger anyway.

test/models/ViewModels/VatCertificateViewModelSpec.scala
```diff
@@ -0,0 +1,54 @@
...
+import org.scalatest.matchers.should.Matchers
+import org.scalatest.wordspec.AnyWordSpecLike
+
+class VatCertificateViewModelSpec extends AnyWordSpecLike with Matchers {
+
+  "The VatCertificateViewModel .fromCustomerInformation()" when {
+
+    lazy val model = VatCertificateViewModel
+
+    "the customer is missing a first or last name" should {
```
lewishay, line 33: Could you also put in a test for the happy path here, i.e. the user has first and last name? I'm guessing it wasn't flagged up in the report due to an unrelated test indirectly covering that branch of code, but it makes sense to have an explicit one for it here.

lewishay (changes requested):

lewishay (approved):
Approved awaiting a clear pipeline

PR #648, opened 2021-12-08
BTAT-9957 Added Welsh messages for penalties banner

lewishay:


JoNB7777 (changes requested):
The changes all look fine. However, I noticed during my review that MockAppConfig is being imported twice into ViewBaseSpec. Could you please remove one of those imports?

JoNB7777 (approved):

PR #658, opened 2021-12-23
[BTAT-10028] Update charges with agent logic

daparkes:


conf/messages
```diff
@@ -212,12 +212,16 @@ chargeType.OADefaultInterestTitle = VAT officer’s assessment interest
...
 chargeType.bnpRegPost2010ChargeDescription = because you should have been registered for VAT earlier
+chargeType.bnpRegPost2010ChargeDescription.agent = because your client should have been registered for VAT earlier
 chargeType.ftnMatPre2010ChargeTitle = Failure to notify penalty
-chargeType.ftnMatPre2010ChargeDescription = you did not tell us you are no longer exempt from VAT registration
+chargeType.ftnMatPre2010ChargeDescription = because you did not tell us you are no longer exempt from VAT registration
+chargeType.ftnMatPre2010ChargeDescription.agent = because your client did not tell us they we were no longer exempt from VAT registration
 chargeType.ftnMatPost2010ChargeTitle = Failure to notify penalty
-chargeType.ftnMatPost2010ChargeDescription = you did not tell us you are no longer exempt from VAT registration
+chargeType.ftnMatPost2010ChargeDescription = because you did not tell us you are no longer exempt from VAT registration
+chargeType.ftnMatPost2010ChargeDescription.agent = because your client did not tell us they we were no longer exempt from VAT registration
```
JoNB7777, line 224: There's a typo in this line (the word 'we' doesn't belong in there). This typo was probably copied over from the main spreadsheet which also contains it. The spreadsheet that only contains the translations doesn't contain this typo though, so we should have the correct translation for the correct sentence. Could you please still fix the typo?

JoNB7777 (changes requested):

conf/messages
```diff
@@ -212,12 +212,16 @@ chargeType.OADefaultInterestTitle = VAT officer’s assessment interest
...
 chargeType.bnpRegPre2010ChargeTitle = Penalty for late registration
 chargeType.bnpRegPre2010ChargeDescription = because you should have been registered for VAT earlier
+chargeType.bnpRegPre2010ChargeDescription.agent = because your client should have been registered for VAT earlier
 chargeType.bnpRegPost2010ChargeTitle = Penalty for late registration
 chargeType.bnpRegPost2010ChargeDescription = because you should have been registered for VAT earlier
+chargeType.bnpRegPost2010ChargeDescription.agent = because your client should have been registered for VAT earlier
 chargeType.ftnMatPre2010ChargeTitle = Failure to notify penalty
-chargeType.ftnMatPre2010ChargeDescription = you did not tell us you are no longer exempt from VAT registration
+chargeType.ftnMatPre2010ChargeDescription = because you did not tell us you are no longer exempt from VAT registration
+chargeType.ftnMatPre2010ChargeDescription.agent = because your client did not tell us they we were no longer exempt from VAT registration
```
JoNB7777, line 221: This line has the same typo as line 224. Could you please fix this as well? Apologies I missed this before

JoNB7777 (changes requested):
I'm also getting lots of scalastyle warnings "Header does not match expected text". We've had these on a few of the other license header tasks as well, I believe in the other cases this was due to the spacing in scalastyle-config.xml. Could you please try to see if changing that fixes it here as well?

JoNB7777 (approved):

PR #663, opened 2022-01-24
[BTAT-10148] Updated charge type logic for open payments page

JoNB7777:


app/views/payments/OpenPayments.scala.html
```diff
@@ -71,7 +71,7 @@ <h1 class="govuk-heading-xl">
       <div class="payments">
         <div class="section">
           @model.payments.zipWithIndex.map { case(payment, count) =>
-            @whatYouOweChargeRow(payment, count, user.isAgent)
+            @whatYouOweChargeRow(payment, count, user.isAgent)(messages, appConfig)
```
lewishay, line 74: No need to pass in implicit params `messages` and `appConfig`

lewishay (changes requested):
Behaviour looks good to me, just a few small changes that can be undone

PR #668, opened 2022-03-10
[BTAT-10322] Create controller for new What You Owe page

daparkes:


test/controllers/WhatYouOweControllerSpec.scala
```diff
@@ -0,0 +1,73 @@
...
+ */
+
+package controllers
+
+import common.TestModels.customerInformationMax
+import org.jsoup.Jsoup
+import play.api.http.Status.OK
+import play.api.test.Helpers._
+
+class WhatYouOweControllerSpec extends ControllerBaseSpec {
```
paulfirmstone, line 24: We will need to add a DD test once the controller has been updated but could you also review the current controller spec to see if we can cover anything else off around auth etc?

paulfirmstone (changes requested):

PR #671, opened 2022-03-15
Btat 10324 Added ChargeTypeDetailsView

JoNB7777:


app/views/payments/ChargeTypeDetailsView.scala.html
```diff
@@ -0,0 +1,130 @@
...
+            ))
+        </div>
+    </div>
+
+    @govukButton(Button(
+        content = Text(messages("chargeTypeDetails.button")),
+        href    = Some(appConfig.unauthenticatedPaymentsUrl)
+    ))
+
+    <a class="govuk-link" href="#"
```
lewishay, line 126: When they aren't wrapped in something else, Govuk links render quite small, and in the screenshot the link looks "standard" size to me. Hard to say for sure without seeing the page here. But I'd play safe and wrap this in a `<p class="govuk-body">`

app/views/payments/ChargeTypeDetailsView.scala.html
```diff
@@ -0,0 +1,130 @@
...
+                        value = Value(
+                            content = HtmlContent(outstandingAmountHtml)
+                        )
+                    )
+                )
+            ))
+        </div>
+    </div>
+
+    @govukButton(Button(
```
lewishay, line 121: Can you add an `if` statement to hide this button for agents

app/views/payments/ChargeTypeDetailsView.scala.html
```diff
@@ -0,0 +1,130 @@
...
+@outstandingAmountHtml = {
+    @model.outstandingAmount
+}
+
+@mainTemplate(
+    title = model.chargeTitle,
+    appConfig = appConfig,
+    serviceInfoContent = serviceInfoContent,
+    user = Some(user),
+    navLinkContent = Some(breadcrumbs)
```
lewishay, line 68: I think we neglected the agent requirements in refinement & subtasking but agents should just have a Back link to WYO, rather than breadcrumbs.

app/views/payments/ChargeTypeDetailsView.scala.html
```diff
@@ -0,0 +1,130 @@
...
+    </div>
+
+    @govukButton(Button(
+        content = Text(messages("chargeTypeDetails.button")),
+        href    = Some(appConfig.unauthenticatedPaymentsUrl)
+    ))
+
+    <a class="govuk-link" href="#"
+        id="whatYouOweLink" rel="noreferrer noopener" target="_blank">
+        @messages("chargeTypeDetails.link")
```
lewishay, line 128: This text is different for agents - "Return to what your client owes". See: https://penalties-prototype.herokuapp.com/beta-baseline/vat-breakdown-m1-agent

test/views/payments/ChargeTypeDetailsViewSpec.scala
```diff
@@ -0,0 +1,192 @@
...
+        }
+
+        "has the correct href" in {
+          element(Selectors.whatYouOweLink).attr("href") shouldBe "#"
+          //TODO: add correct link location once this page and the What you owe page have been wired up
+        }
+      }
+    }
+
+    "the user has a cleared amount and no period for the charge" should {
```
lewishay, line 164: Can you add in another scenario for when the user only has one period date? As a way to verify your && logic that requires both to render the caption.

lewishay (changes requested):

PR #673, opened 2022-03-17
BTAT-10323 what you owe wiring up

SheilaTran:


app/testOnly/controllers/WhatYouOweController.scala
```diff
@@ -0,0 +1,68 @@
...
+            periodKey = Some("#001"),
+            isOverdue = false,
+            chargeReference = Some("XD002750002155"),
+            makePaymentRedirect = "/vat-through-software/make-payment",
+            periodFrom = Some(LocalDate.parse("2017-01-01")),
+            periodTo = Some(LocalDate.parse("2017-03-01"))
+          ))
+        )
+
+        Future.successful(Ok(whatYouOwe(viewModel, serviceInfoContent)))
```
lewishay, line 64: Can you change the above service call to use `map` instead of `flatMap` and then remove the `Future` wrapping from here

app/testOnly/controllers/WhatYouOweController.scala
```diff
@@ -0,0 +1,68 @@
...
+  implicit val ec: ExecutionContext = mcc.executionContext
+
+  def show: Action[AnyContent] = authorisedController.financialAction { implicit request =>
+    implicit user => ddInterrupt.interruptCheck { _ =>
+
+      serviceInfoService.getPartial.flatMap { serviceInfoContent =>
+
+        val viewModel = WhatYouOweViewModel(
+          totalAmount = 1000.00,
+          charges = Seq(WhatYouOweChargeModel(
```
lewishay, line 48: Could you add an additional charge to this list - ideally an overdue one - and update the total amount to match. Then we should have a fuller view of how things are rendering on the page (overdue label, multiple items in the table)

lewishay (changes requested):

PR #681, opened 2022-03-29
[BTAT-10351-a] Handle missing charge type description

daparkes:


lewishay:
Commit was merged to main but the PR has remained open - looks to be a glitch

PR #683, opened 2022-04-01
[BTAT-10365] Wire up WhatYouOwe and charge detail views

daparkes:


app/views/payments/WhatYouOwe.scala.html
```diff
@@ -67,7 +68,22 @@
   @if(charge.isOverdue){@govukTag(Tag(
     content = Text(messages("common.overdue")),
     classes = "govuk-tag--red"
-  ))}<a class="govuk-link" href="/todo-new-page">@charge.chargeTitle @charge.chargeDescription</a>
+  ))}
+  @form(action = testOnly.controllers.routes.ChargeBreakdownController.showBreakdown) {
```
lewishay, line 72: The overdue label is now rendering above the link rather than next to it; can we change that? 
image

lewishay (changes requested):

app/assets/stylesheets/application.scss
```diff
@@ -34,3 +34,23 @@ dl.govuk-\!-margin-bottom-0 > .govuk-summary-list__row > .govuk-summary-list__ke
     border-bottom: 1px solid $govuk-border-colour;
   }
 }
+
+.govuk-table__cell {
```
lewishay, line 38: This has messed with the payment history view as it also uses a govuk table. Can you use a custom class for these changes to ensure this page the only thing being affected

lewishay (changes requested):

app/views/payments/WhatYouOwe.scala.html
lewishay: I'm still seeing it with the new changes. The alignment is now also a bit off. This is on Firefox.
image

lewishay (changes requested):

PR #690, opened 2022-04-26
Btat 10414 Added new field to payment history and subitem models

JoNB7777:


lewishay (approved):
Awaiting clear pipeline

PR #694, opened 2022-05-05
[BTAT-10441] - Added vat return 1st lpp in charge type model

william-hatzar:


paulfirmstone (changes requested):
I tried to load the pages following the new stub data being applied and we get an error. It looks like we need to widen the scope of this task to also add the new charge type to PaymentMessagesHelper in order for the page to render and the AT's to pass. Can you look to update any tests specific to the charge type if there are any.

paulfirmstone (changes requested):
Although you haven't changed the files directly there are additional tests we can add for the new charge type. Could you add to the following specs and check for any others:

MakePaymentControllerSpec
WhatYouOweChargerhelperSpec 
PaymentsHistoryChargeDescrioptionTemplateSpec
PaymentsHttpParserSpec

PR #696, opened 2022-05-05
BTAT-10448 rename components

SheilaTran:
I've renamed InterestCharge to EstimatedInterest, but I am also open to suggestions 😊

### DIFF
--- a/app/testOnly/controllers/ChargeBreakdownController.scala
+++ b/app/testOnly/controllers/ChargeBreakdownController.scala
@@ -35,7 +35,7 @@ class ChargeBreakdownController @Inject()(authorisedController: AuthorisedContro
                                           mcc: MessagesControllerComponents,
                                           serviceInfoService: ServiceInfoService,
                                           chargeBreakdownView: ChargeTypeDetailsView,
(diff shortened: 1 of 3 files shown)


app/testOnly/controllers/ChargeBreakdownController.scala
```diff
@@ -61,7 +60,7 @@ class ChargeBreakdownController @Inject()(authorisedController: AuthorisedContro
     implicit user => DDInterrupt.interruptCheck { _ =>
       if(appConfig.features.interestBreakdownEnabled()) {
         serviceInfoService.getPartial.map { navLinks =>
-          InterestChargeViewModel.form.bindFromRequest.fold(
+          EstimatedInterestViewModel.form.bindFromRequest.fold(
```
lewishay, line 63: Could you please also rename the above controller action, and the associated route name (placeholder for now). I did the task recently to add the new action and I called it `crystallisedInterestBreakdown`, with the route being `/crystallised-interest` so I think it makes sense to follow that pattern here too

lewishay (changes requested):

PR #709, opened 2022-05-31
[BTAT-10535] Update bootstrap version

daparkes:


it/pages/LanguageSpec.scala
```diff
@@ -69,7 +66,7 @@ class LanguageSpec extends IntegrationBaseSpec {
 
     "language preference is not set in the cookie" should {
```
lewishay, line 67: This test is now incorrect as the language is being set to english by default in the cookie. May be best remedied by having lang as optional string and a small bit of logic in the cookie baker?

lewishay (changes requested):

PR #714, opened 2022-06-16
BTAT-10617 fix tile to exclude PaymentOnAccount

SheilaTran:


app/services/VatDetailsService.scala
```diff
@@ -51,7 +51,7 @@ class VatDetailsService @Inject()(vatObligationsConnector: VatObligationsConnect
 
     financialDataConnector.getOpenPayments(vrn).map {
       case Right(payments) =>
-        val outstandingPayments = payments.financialTransactions.filter(_.outstandingAmount > 0)
+        val outstandingPayments = payments.financialTransactions.filter(_.outstandingAmount > 0).filterNot(_.chargeType equals PaymentOnAccount)
```
daparkes, line 54: Could you trim down this line a bit, e.g. by putting each .filter on a new line

daparkes (changes requested):
Could you also add a unit test in the service spec to check that payments on account are filtered out? It would be nice to check that payments with outstandingAmount = 0 are being filtered out too, so maybe you could pass in both kinds of payment in one test, along with a normal payment, and make sure only the normal payment is returned

PR #716, opened 2022-06-16
[BTAT-10597] Removed reference to number of days in estimated interes…

JoNB7777:
…t view and view model


paulfirmstone (changes requested):
This looks good but there is one reference to numberOfDaysLate in the ChargeBreakdownControllerSpec. Can you remove it please?

PR #722, opened 2022-06-23
[BTAT-10603] - Updated inconsistencies and also simplified date helper

william-hatzar:


lewishay (changes requested):

lewishay:
AT here: https://github.com/hmrc/view-vat-acceptance-tests/pull/312
Although the ATs still needs the above requested changes to this PR before they will pass

PR #733, opened 2022-07-07
Refactor to WYO form/button solution

lewishay:


app/views/templates/payments/wyoCharges/StandardCharge.scala.html
```diff
@@ -18,34 +18,22 @@
 @import models.viewModels.StandardChargeViewModel
 @import views.templates.formatters.dates.DisplayDateRangeHelper.displayDate
 
-@this(form: FormWithCSRF, govukTag: GovukTag)
+@this(govukTag: GovukTag)
 
 @(charge: StandardChargeViewModel)(implicit request: Request[_], messages: Messages, appConfig: AppConfig, user: User)
```
william-hatzar, line 23: When running this locally, a warning appears saying that the request param in the header is unused so I don't think we need this. Could you remove it? Also, although not the scope of the task we seem to have a few other unused imports that have appeared in the warning logs I believe from my task that went in for the crystallised interest case. I'm not sure why it's complaining now as they didn't appear when I did my task, but one is in CrystallisedCharge.scala and is complaining of an unused AppConfig import and the other is the PaymentMessageHelper import in the crystalledInterestView.scala. Other than that the code looks good to me I have run the unit tests and also tested it locally and can't seem to see anything wrong with it 😊.

william-hatzar (approved):

PR #738, opened 2022-07-15
[BTAT-10645] Refactor categoriseCharges; add interestAmount to view m…

daparkes:
…odel


test/common/TestModels.scala
```diff
@@ -433,4 +433,64 @@ object TestModels {
   val whatYouOweUrl: String = testOnly.controllers.routes.WhatYouOweController.show.url
 
   val vatDetailsUrl: String = controllers.routes.VatDetailsController.details.url
+
+  val nonInterestChargeTypes: Seq[ChargeType] = Vector(
```
lewishay, line 437: To make this more concise and maintainable, can you:

- Change the interest & penalty charge type lists in `ChargeType` to be Sets of type `Set[ChargeType]` (should have been this anyway tbh to match `allChargeTypes`)
- Update the val here to simply be `allChargeTypes.diff(interestChargeTypes)` (or preferably if this is only being used in one unit test, move it there)

lewishay (changes requested):

lewishay (approved):
Looks good to me now, confirmed charges are in the correct order on the page & interest amount value correctly passed down from API. Awaiting clear pipeline

PR #740, opened 2022-07-18
BTAT-10711 model for the new mongo repository

SheilaTran:


william-hatzar (approved):

william-hatzar (commented):
There is an unused import in the charge types file for penaltyChargeTypes. Could you remove this as I don't think it's needed

PR #741, opened 2022-07-19
BTAT 10706 - Added model, connector and parser for 1812 call

william-hatzar:


app/models/penalties/LPPDetails.scala
```diff
@@ -0,0 +1,37 @@
...
+                      LPP1HRCalculationAmount: Option[BigDecimal],
+                      LPP1HRDays: Option[String],
+                      LPP1HRPercentage: Option[Double],
+                      LPP2Days: Option[String],
+                      LPP2Percentage: Option[Double],
+                      penaltyChargeReference: Option[String])
+
+object LPPDetails {
+
+  implicit val format: Format[LPPDetails] = Json.format[LPPDetails]
```
lewishay, line 35: Can you change this to be a JSON reads instead of a format. We don't need to write this model to JSON, and it will save you testing it

app/models/penalties/PenaltyDetails.scala
```diff
@@ -0,0 +1,27 @@
...
+
+package models.penalties
+
+import play.api.libs.json.{Format, Json}
+
+case class PenaltyDetails(LPPDetails : Seq[LPPDetails])
+
+object PenaltyDetails {
+
+  implicit val format: Format[PenaltyDetails] = Json.format[PenaltyDetails]
```
lewishay, line 25: Can you change this to be a JSON reads instead of a format. We don't need to write this model to JSON, and it will save you testing it

test/models/penalties/LPPDetailsSpec.scala
```diff
@@ -0,0 +1,53 @@
...
+
+      "optional fields are present" in {
+        LPPDetailsJsonMax.as[LPPDetails] shouldBe LPPDetailsModelMax
+      }
+      "optional are not present" in {
+        LPPDetailsJsonMin.as[LPPDetails] shouldBe LPPDetailsModelMin
+      }
+    }
+
+    "serialize to json" when {
```
lewishay, line 40: as mentioned above, no need for JSON writing functionality or tests

test/models/penalties/PenaltyDetailsSpec.scala
```diff
@@ -0,0 +1,53 @@
...
+      "optional fields are present" in {
+        penaltyDetailsJsonMax.as[PenaltyDetails] shouldBe penaltyDetailsModelMax
+      }
+
+      "optional fields are not present" in {
+       penaltyDetailsJsonMin.as[PenaltyDetails] shouldBe penaltyDetailsModelMin
+      }
+    }
+
+    "serialize to json" when {
```
lewishay, line 39: as mentioned above, no need for JSON writing functionality or tests

lewishay (changes requested):

it/stubs/PenaltyDetailsStub.scala
```diff
@@ -0,0 +1,90 @@
...
+    Some("31"),
+    Some(5.5),
+    penaltyChargeReference = Some("BCDEFGHIJKLMNOPQ")
+  )
+
+  val penaltyDetailsModelMax: PenaltyDetails = PenaltyDetails(
+    LPPDetails = Seq(LPPDetailsModelMax)
+  )
+
+  val LPPDetailsModelMin: LPPDetails = LPPDetails(
```
lewishay, line 52: There's still an unused model here

test/connectors/httpParsers/PenaltyDetailsHttpParserSpec.scala
```diff
@@ -0,0 +1,72 @@
...
+      }
+    }
+
+    "the http response status is 404 NOT_FOUND no LPP data is returned" should {
+
+      val httpResponse = HttpResponse(Status.NOT_FOUND, "")
+      val expected = Right(PenaltyDetails(List()))
+      val result = PenaltyDetailsReads.read("", "", httpResponse)
+
+      "return a 404 error" in {
```
lewishay, line 51: I missed this the first time around - but the description here is wrong, the parser doesn't return a 404 error for this scenario, it returns a `PenaltyDetails` object with an empty sequence

lewishay (changes requested):

PR #749, opened 2022-07-26
BTAT-10713 Added WYO session service to speak to repository layer

lewishay:
- I have added Mockito dependency as ScalaMock does not play nice with mocking out mongo repositories
- I had to add the reptitive `Json.toJson` line into the write function as there is no JSON writes for the generic type
- I elected to keep the read function simple considering the controller action will need to have a match statement for the view model type again anyway - it seemed pointless to duplicate it
- The tests are quite simple as they are essentially just returning what the database does (extra scenarios would simply be - when I mock a `false`, return `false` - when I mock a `None`, return `None` - etc)

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -52,11 +52,11 @@ val compile: Seq[ModuleID] = Seq(
 )
 
 def test(scope: String = "test, it"): Seq[ModuleID] = Seq(
-  "uk.gov.hmrc"             %% "bootstrap-test-play-28"      % "5.24.0",
-  "org.pegdown"             %  "pegdown"                     % "1.6.0"      % scope,
-  "org.jsoup"               %  "jsoup"                       % "1.14.3"     % scope,
-  "org.scalamock"           %% "scalamock-scalatest-support" % "3.6.0"      % scope,
-  "uk.gov.hmrc.mongo"       %% "hmrc-mongo-test-play-28"     % "0.67.0"    % scope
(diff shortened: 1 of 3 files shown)


build.sbt
```diff
@@ -52,11 +52,11 @@ val compile: Seq[ModuleID] = Seq(
...
-  "uk.gov.hmrc"             %% "bootstrap-test-play-28"      % "5.24.0",
-  "org.pegdown"             %  "pegdown"                     % "1.6.0"      % scope,
-  "org.jsoup"               %  "jsoup"                       % "1.14.3"     % scope,
-  "org.scalamock"           %% "scalamock-scalatest-support" % "3.6.0"      % scope,
-  "uk.gov.hmrc.mongo"       %% "hmrc-mongo-test-play-28"     % "0.67.0"    % scope
+  "uk.gov.hmrc"        %% "bootstrap-test-play-28"      % "5.24.0"  % scope,
+  "org.scalatestplus"  %% "mockito-3-4"                 % "3.2.9.0" % scope,
+  "org.jsoup"          %  "jsoup"                       % "1.14.3"  % scope,
+  "org.scalamock"      %% "scalamock-scalatest-support" % "3.6.0"   % scope,
+  "uk.gov.hmrc.mongo"  %% "hmrc-mongo-test-play-28"     % "0.67.0"  % scope
```
paulfirmstone, line 59: Can you bump this and the main dependency to version 0.68.0 so we don't hit the Bobby rule.

paulfirmstone (changes requested):

PR #751, opened 2022-07-29
[BTAT-10735] Remove charge type filtering

daparkes:
This can be reviewed but cannot be merged until filtering is added to financial-transactions


lewishay (changes requested):

lewishay (approved):
Approved awaiting other tasks

PR #752, opened 2022-08-01
[BTAT-10738] Updated Crystallised LPP1 parameters

JoNB7777:


daparkes (changes requested):
The story description was updated on Thursday to include a couple of other things. Could you please:
- split the interestRate field into 2 parts
- update the view to read them separately

daparkes (approved):

PR #754, opened 2022-08-02
[BTAT-10726] - Added mongo writes and finished tests

william-hatzar:


test/services/PenaltiesServiceSpec.scala
```diff
@@ -44,11 +50,25 @@ class PenaltiesServiceSpec extends AnyWordSpecLike with MockFactory with Matcher
     new PenaltiesService(mockPenaltiesConnector)
   }
 
-  "Calling getPenaltiesDataForVRN" should {
+  def penaltiesServiceWithoutMock() : PenaltiesService = {
```
daparkes, line 53: Sorry to nitpick, but you don't need these 2 versions of the service. You can just have `def penaltiesService(): PenaltiesService = new PenaltiesService(mockPenaltiesConnector)` and then call `setup()` only in the test that needs the mock

app/testOnly/controllers/WhatYouOweController.scala
```diff
@@ -21,12 +21,13 @@ import common.SessionKeys
 import config.AppConfig
 import controllers.AuthorisedController
 import controllers.predicates.DDInterruptPredicate
+import models.{User, WYODatabaseModel}
 import models.errors.PenaltiesFeatureSwitchError
```
daparkes, line 25: There are some unused imports here

PR #757, opened 2022-08-03
BTAT-10743 update charge breakdown pages and fix bug

SheilaTran:


paulfirmstone (changes requested):
Can you also move the caption into the H1 for the open and no payment views where we add the client name for Agents.

SheilaTran:
AT PR: https://github.com/hmrc/change-vat-details-acceptance-tests/pull/202

paulfirmstone (approved):

PR #767, opened 2022-08-16
[BTAT-10656] Add AA LPP LPI charge types

daparkes:
https://github.com/hmrc/financial-transactions/pull/82

### DIFF
--- a/app/models/payments/ChargeType.scala
+++ b/app/models/payments/ChargeType.scala
@@ -219,27 +219,27 @@ case object VatReturn2ndLPPLPI extends ChargeType {
 case object VatCentralAssessmentLPI extends ChargeType {
   override val value: String = "VAT Central Assessment LPI"
 }
-
 case object VatCA1stLPPLPI extends ChargeType {
   override val value: String = "VAT CA 1st LPP LPI"
 }
-
 case object VatCA2ndLPPLPI extends ChargeType {
   override val value: String = "VAT CA 2nd LPP LPI"
 }
-
 case object VatOfficersAssessmentLPI extends ChargeType {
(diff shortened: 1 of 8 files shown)


daparkes:
One of the charge breakdown view models had a `description` method, which we only need for the charge rows and not the breakdowns, so I got rid of it

test/views/templates/payments/PaymentsHistoryChargeDescriptionTemplateSpec.scala
```diff
@@ -1467,5 +1497,36 @@ class PaymentsHistoryChargeDescriptionTemplateSpec extends ViewBaseSpec {
         elementText(Selectors.description) shouldBe "for late payment of protective assessment for 12 Jan to 23 Mar 2018"
       }
     }
+
+    "there is a VAT AA 1st LPP LPI charge" should {
+
+      val model = exampleModel.copy(chargeType = VatOA2ndLPPLPI)
```
paulfirmstone, line 1503: This is testing the wrong charge type.

app/models/payments/ChargeType.scala
```diff
@@ -374,7 +373,9 @@ object ChargeType extends LoggerUtil {
     VatOA1stLPPLPI,
     VatOA2ndLPPLPI,
     VatPA1stLPPLPI,
-    VatPA2ndLPPLPI
+    VatPA2ndLPPLPI,
```
paulfirmstone, line 376: These charges also need to be in interestChargeTypes in order to generate the correct breakdown screen.

paulfirmstone (changes requested):

PR #769, opened 2022-08-16
[BTAT-10768] Removed timer from connectors

JoNB7777:


william-hatzar:
Hey, I've just had a look at your PR, you can remove the whole metric service along with the failure counters in the connectors as we can't see those either in grafana.

PR #772, opened 2022-08-22
Btat 10763

JoNB7777:
Please review together with https://github.com/hmrc/financial-transactions/pull/87


william-hatzar (commented):
Could you also add some tests to the WhatYouOweHelperSpec. These also have agent tests so need to be added there as well. Thanks

PR #790, opened 2022-09-21
[BTAT-10869] Added description and title to ChargeDetailsViewModel trait

JoNB7777:


app/models/viewModels/ChargeDetailsViewModel.scala
```diff
@@ -16,10 +16,19 @@
...
+import views.templates.payments.PaymentMessageHelper
+
 import java.security.MessageDigest
 
 trait ChargeDetailsViewModel {
+  val chargeType: String
   val outstandingAmount: BigDecimal
+  val overdue: Boolean = false
+
+  def title(implicit messages: Messages): String = messages(PaymentMessageHelper.getChargeType(chargeType).title)
```
lewishay, line 29: Could you move the tests around for this? It's currently being tested in a few of the subclass view model specs (looks to be all the LPP models?) but I think following this change it makes more sense to remove those repeats and just have one test for it in `ChargeDetailsViewModelSpec`.

lewishay (changes requested):

PR #795, opened 2022-09-29
BTAT-10849 LSP view and view model

SheilaTran:


app/models/viewModels/LateSubmissionPenaltyViewModel.scala
```diff
@@ -0,0 +1,59 @@
...
+
+package models.viewModels
+
+import play.api.i18n.Messages
+import play.api.libs.json.{Json, OFormat}
+import views.templates.payments.PaymentMessageHelper
+
+import java.time.LocalDate
+
+case class LateSubmissionPenaltyViewModel(chargeType: String,
```
lewishay, line 25: Can you add a test spec for this view model please. Looks as though a good range of tests could target: the `description` function, the `makePaymentRedirect` val, and the JSON reads/writes.

Could you also add one test case to the existing set of tests in `ChargeDetailsViewModelSpec` that test the `title` function. The function is well tested already but seeing as we have one for every other type of view model, might as well keep that exhaustive.

app/models/viewModels/LateSubmissionPenaltyViewModel.scala
```diff
@@ -0,0 +1,59 @@
...
+                                          dueDate: LocalDate,
+                                          penaltyAmount: BigDecimal,
+                                          amountReceived: BigDecimal,
+                                          leftToPay: BigDecimal,
+                                          isOverdue: Boolean,
+                                          chargeReference: String,
+                                          periodFrom: LocalDate,
+                                          periodTo: LocalDate) extends ChargeDetailsViewModel {
```
lewishay, line 33: This model needs to extend trait `CrystallisedViewModel` in order for the appropriate WYO row to render

test/views/payments/LateSubmissionPenaltyViewSpec.scala
```diff
@@ -0,0 +1,200 @@
...
+      "has the correct link text" in {
+        elementText("#content > div > div > p:nth-child(4) > a") shouldBe "Return to what your client owes"
+      }
+
+      "has the correct href" in {
+        element("#content > div > div > p:nth-child(4) > a").attr("href") shouldBe whatYouOweLink
+      }
+    }
+
+    "not render breadcrumbs" in {
```
lewishay, line 181: Considering you're testing when breadcrumbs/pay button are not visible for an agent, does it make sense to throw in a test for the principal user to confirm back link is not visible?

PR #805, opened 2022-10-10
BTAT-10924 update WYO controller

SheilaTran:


test/testOnly/controllers/WhatYouOweControllerSpec.scala
```diff
@@ -652,6 +653,26 @@ class WhatYouOweControllerSpec extends ControllerBaseSpec {
       }
     }
 
+    "return a CrystallisedLPP2ViewModel" when {
+
+      "originalAmount, chargeReference and all appropriate LPP2 penalty details are present" in {
```
lewishay, line 658: `originalAmount` is no longer optional so doesn't play a part in the logic here.

Related to this I can see a number of places in this test spec that are still referring to `originalAmount` being present, including a strangely commented out test case on line 477. Can these all be addressed here too please.

lewishay (changes requested):

PR #811, opened 2022-10-28
[BTAT-10972] Added hidden text to link on WYO page

JoNB7777:


app/views/templates/payments/wyoCharges/StandardCharge.scala.html
```diff
@@ -44,10 +44,11 @@
 <br>
 
 <span class="govuk-hint govuk-!-margin-0">
-  @messages("whatYouOwe.due") @displayDate(charge.dueDate)
+  <span id="what-you-owe-due-date">@messages("whatYouOwe.due") @displayDate(charge.dueDate)</span>
   @if(charge.viewReturnEnabled && charge.periodKey.isDefined) {
     <a class="view-return-link govuk-link govuk-!-padding-left-1" href="@viewReturnHref">
-      @messages("whatYouOwe.viewReturn")
+      <span id="what-you-owe-view-return">@messages("whatYouOwe.viewReturn")</span>
+      <span id="what-you-owe-view-return-hidden-text" class="govuk-visually-hidden">@messages("whatYouOwe.viewReturn") @messages("openPayments.vatReturn", displayDateRange(charge.periodFrom.get, charge.periodTo.get))</span>
```
daparkes, line 51: Unfortunately this crashes the whole page with an exception if there's no period info because of `NoSuchElementException: None.get`. We need a case match in here or similar to handle that, maybe just not rendering the hidden text if the fields aren't there. The scenario shouldn't happen anyway, but I can't see any actual logic that would guard against it at the moment

daparkes (changes requested):

PR #817, opened 2022-11-17
BTAT-11020 Payment reads for interest rate

lewishay:
Also removed lots of bloat from `Payment` model, and refactored the tests for that model. The many `apply` functions were only being used across various test specs and this was unnecessary. This unfortunately meant making the same change to many test models which clogs up the PR a bit.

Have not been able to manually or acceptance test this yet without stub data


daparkes (approved):

PR #819, opened 2022-11-18
BTAT-11034 new penalties banner

SheilaTran:
Still missing welsh, but the rest can be reviewed

### DIFF
--- a/app/views/templates/penalties/PenaltiesBanner.scala.html
+++ b/app/views/templates/penalties/PenaltiesBanner.scala.html
@@ -45,10 +45,32 @@
   </a>
 }
 
+@penaltiesChangesBannerContent = {
+
+<div class="govuk-!-margin-bottom-3">
+
+    <h3 class="govuk-notification-banner__heading">@messages("penaltiesChangesBanner.heading")</h3>
+
+    <p class="govuk-body" id="announcement-information">@messages("penaltiesChangesBanner.message.1")</p>
+    <p class="govuk-body" id="date-information">@messages("penaltiesChangesBanner.message.2")</p>
+    <p class="govuk-body" id="calculation-information">@messages("penaltiesChangesBanner.message.3")</p>
+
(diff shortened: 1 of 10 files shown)


app/views/templates/penalties/PenaltiesBanner.scala.html
```diff
@@ -45,10 +45,32 @@
...
+</div>
+}
+
 @if(penaltiesSummary.hasActivePenalties) {
   @govukNotificationBanner(NotificationBanner(
     title = Text(messages("penaltiesBanner.heading")),
     content = HtmlContent(bannerContent),
     classes = "govuk-!-margin-top-6"
   ))
+} else {
```
william-hatzar, line 71: You need to have another condition whereby if the feature switch for the penalties banner has been turned off then this banner still shows. If I turn off the feature switch then this banner disappears. This is why we need the welsh for the story.

william-hatzar (dismissed):

test/views/vatDetails/VatDetailsViewSpec.scala
```diff
@@ -522,24 +522,22 @@ class VatDetailsViewSpec extends ViewBaseSpec {
...
           "has the correct text" in {
-            penaltiesBanner.select("a").text shouldBe "Find out why you have penalties"
+            elementText(".govuk-notification-banner__link") shouldBe "Find out why you have penalties"
           }
 
           "has the correct link destination" in {
-            penaltiesBanner.select("a").attr("href") shouldBe mockConfig.penaltiesFrontendUrl
+            element(".govuk-notification-banner__link").attr("href") shouldBe mockConfig.penaltiesFrontendUrl
           }
         }
```
william-hatzar, line 542: In this testSpec there are some tests where the Penalties Banner isn't showed due to no penalty points. I think these could be refactored so that instead it shows the penalties are coming banner instead of nothing.

william-hatzar (approved):

PR #828, opened 2022-11-28
BTAT-11052 Updated message Interest on VAT

lz1kwn:


paulfirmstone (changes requested):
There is an agent version of this content. Can you create a message file entry and adjust the view so it handles both. See 'estimatedLPP2.estimatesDescriptionAgent' for the content. The estimated LPP2 view will also have an example of how we determine if it is an agent or trader accessing the view.

conf/messages
```diff
@@ -179,7 +180,8 @@ estimatedLPP1.amountReceived = Amount received
 estimatedLPP1.leftToPay = Left to pay
 estimatedLPP1.warning = The penalty will increase by a further {0}% of the unpaid VAT, if VAT remains unpaid {1} days after the due date.
 estimatedLPP1.estimatesHeading = Estimates
-estimatedLPP1.estimatesSubHeading = Penalties will show as estimates until:
+estimatedLPP1.estimatesSubHeading = Penalties will show as estimates until you pay the charge they relate to.
```
paulfirmstone, line 183: You shouldn't be changing anything to do with LPP1 or LPP2. This task is for estimated interest only.

conf/messages
```diff
@@ -168,7 +168,8 @@ estimatedInterest.currentAmount = Current amount (estimate)
 estimatedInterest.amountReceived = Amount received
 estimatedInterest.leftToPay = Left to pay
 estimatedInterest.estimates = Estimates
-estimatedInterest.estimatesExplanation = Penalties and interest will show as estimates if HMRC has not been given enough information to calculate the final amounts.
+estimatedInterest.estimatesExplanation = Penalties and interest will show as estimates until you pay the charge they relate to.
+estimatedInterest.estimatesDescriptionAgent = Penalties and interest will show as estimates until your client pays the charge they relate to.
```
paulfirmstone, line 172: Can you create an entry in the same place for this in the messages.cy file.

paulfirmstone (changes requested):

conf/messages
lz1kwn: Reverted back to original code

conf/messages
lz1kwn: Added entry in messages cy

PR #831, opened 2022-12-02
[BTAT-11105] Add OA and CA LPP charge types

daparkes:


app/views/templates/payments/PaymentMessageHelper.scala
```diff
@@ -826,7 +854,11 @@ object PaymentMessageHelper {
     VATPOAReturn1stLPP,
     VATPOAReturn2ndLPP,
     VATReturnPOA1stLPPLPI,
-    VATReturnPOA2ndLPPLPI
+    VATReturnPOA2ndLPPLPI,
+    VATReturnCA1stLPP,
```
paulfirmstone, line 858: Do the central assessments need to have the word return in them? Can these be VATCA1stLPP etc?

paulfirmstone (changes requested):

paulfirmstone (approved):

PR #840, opened 2023-01-12
[BTAT-11245] - Update link penalties changing banner

william-hatzar:


conf/application.conf
```diff
@@ -282,7 +282,7 @@ gov-uk {
   vatCorrections = "https://www.gov.uk/vat-corrections"
   mtdGuidance = "https://www.gov.uk/guidance/when-to-start-using-making-tax-digital-for-vat-if-youve-not-before"
   govUkHolding = "https://www.gov.uk"
-  penaltiesChanges = "https://www.gov.uk/guidance/prepare-for-upcoming-changes-to-vat-penalties-and-vat-interest-charges"
+  penaltiesChanges = "https://www.gov.uk/government/collections/vat-penalties-and-interest"
```
paulfirmstone, line 285: Doesn't necessarily effect the test itself but we do check for the old URL in the spec. Can you update it to be consistent?

paulfirmstone (changes requested):

PR #846, opened 2023-01-26
BTAT-11290 Added support for breathingSpace boolean

lewishay:
I added in multiple tests to make sure an empty penalties array with breathing space boolean is supported, as we no longer want to treat no penalty details as 404. A change will be made in the backend to accomodate for this. Please don't merge until the backend work is ready.


lewishay:
> * `Scala` can be updated from `2.13.8` to `2.13.10` in [build.sbt](https://github.com/hmrc/vat-summary-frontend/blob/BTAT-11290/build.sbt#L89)

We've still not seen a way to achieve this without error, I'm leaving it for now but I'll create a backlog ticket for investigation.

PR #872, opened 2023-03-24
[BTAT-11451] Correct Payment History POA logic

daparkes:


lewishay (approved):
Awaiting stub data approval

PR #877, opened 2023-04-19
BTAT-11523 Adding audit event for WYO Pay Now button

paulfirmstone:


app/controllers/MakePaymentController.scala
```diff
@@ -125,4 +126,13 @@ class MakePaymentController @Inject()(paymentsService: PaymentsService,
         InternalServerError(paymentsError())
     }
   }
+
+  def makeFullPaymentHandoff: Action[AnyContent] =
+    authorisedController.authorisedAction { implicit request => user => {
+
+        auditingService.audit(PayFullChargeAuditModel(user))
```
lewishay, line 133: Can you provide the `path` parameter in this audit function

conf/app.routes
```diff
@@ -8,6 +8,7 @@ GET        /vat-certificate
 
 GET        /make-payment/:amountInPence/:taxPeriodMonth/:taxPeriodYear/:vatPeriodEnding/:chargeType/:dueDate/:chargeReference        controllers.MakePaymentController.makePayment(amountInPence: Long, taxPeriodMonth: Int, taxPeriodYear: Int, vatPeriodEnding: String, chargeType: String, dueDate: String, chargeReference: String)
 GET        /make-payment/:amountInPence/:chargeType/:dueDate/:chargeReference                                                        controllers.MakePaymentController.makePaymentNoPeriod(amountInPence: Long, chargeType: String, dueDate: String, chargeReference: String)
+GET        /pay-api/view-and-change/vat/                                                                                             controllers.MakePaymentController.makeFullPaymentHandoff
```
lewishay, line 11: Any reason why you chose this route name? It's a little misleading as `pay-api` is the context URL for the microservice `pay-api`. Could we instead just use something like `/make-payment` so it's similar to our other payment URLs above?

test/controllers/MakePaymentControllerSpec.scala
```diff
@@ -328,4 +328,77 @@ class MakePaymentControllerSpec extends ControllerBaseSpec {
...
+
+      "return status 403(Forbiddden)" in {
+        status(result) shouldBe Status.FORBIDDEN
+      }
+    }
+
+    "the user is an Agent" should {
+      lazy val result = {
+        mockAgentAuth()
+        mockAudit()
```
lewishay, line 392: According to the code there shouldn't be an audit if the user is an agent. Can you remove this mock to confirm. Perhaps Scalamock doesn't flag it if you set a mock up and it isn't called, so it didn't cause an issue

lewishay (changes requested):

lewishay:
test this please

lewishay:
test this please

PR #879, opened 2023-04-25
BTAT-11545 Upgraded dependencies

lewishay:
Also requires this AT change: https://github.com/hmrc/view-vat-acceptance-tests/pull/346

I decided to remove `RoutesSpec` as it felt unnecessary to be proving Play reverse routes functionality moreso than other unit tests already are, and it wasn't comprehensive of all service routes anyway. Let me know if you disagree.


paulfirmstone (approved):
Approved but blocked behind pipeline

PR #912, opened 2024-01-26
DL-12343 Implement VAT Return POA RPI

arunankannan:


app/models/payments/ChargeType.scala
```diff
@@ -242,6 +242,9 @@ case object VatReturnPOA1stLPPLPI extends ChargeType {
 case object VatReturnPOA2ndLPPLPI extends ChargeType {
   override val value: String = "VAT Return POA 2nd LPP LPI"
 }
+case object VatReturnPOARPI extends ChargeType {
```
naglisr, line 245: Document the parent charge type here. For an example, see https://github.com/hmrc/vat-summary-frontend/blob/a61151bec7cf1b661dede1395385d1f11f5bee22/app/models/payments/ChargeType.scala#L400

PR #922, opened 2024-02-22
DL-12374 added manual and 1st return LPP RPI's

RoheelArshad:


test/views/templates/payments/PaymentsHistoryChargeDescriptionTemplateSpec.scala
```diff
@@ -2360,6 +2361,62 @@ class PaymentsHistoryChargeDescriptionTemplateSpec extends ViewBaseSpec {
         elementText(Selectors.description) shouldBe "for period 12\u00a0Jan to 23\u00a0Mar\u00a02018"
       }
     }
+
```
Thepworth95, line 2364: All Tests including non agent tests are in the `"user is an agent" when {` part. can the non agent tests be put with the rest of the tests to keep them organise

RoheelArshad: Cool updated now

Thepworth95 (approved):

PR #950, opened 2024-05-14
[DL-13719] Updating agent content previously referencing classic serv…

JoNB7777:
…ices


JoNB7777:
These changes were included in https://github.com/hmrc/vat-summary-frontend/pull/949 instead, so this PR isn't needed anymore

PR #957, opened 2025-02-13
DL-15196 Interstitial page for Traders with DD

shankaran-hmrc:


app/controllers/ExistingDirectDebitController.scala
```diff
@@ -0,0 +1,97 @@
...
+                            earliestDueDate = formModel.dueDateOrUrl,
+                            linkId = "existing-dd-pay-now-button"
+                          ).url
+                          infoLog(s"[ExistingDirectDebitController] [submit] " +
+                            s"User clicked Yes to pay even DD has hence navigating to payment " + makePaymentRedirect)
+                           Redirect(makePaymentRedirect)
+                        case _ =>
+                          infoLog(s" [ExistingDirectDebitController] [submit] " +
+                            s"User clicked Yes to pay even DD has hence navigating to payment " + formModel.dueDateOrUrl.get)
+                          Redirect(formModel.dueDateOrUrl.get)
```
shankaran-hmrc, line 83: Hi @ahlee0 RedirectUrl object is suggested by plat-op is applicable for immediate continue redirect urls, callbacks url only but in our case we are submitting form with /direct-debit-yes url and then doing validation in backend and then redirecting to url from query param. I have implemented RedirectUrl but when we submit form its failing with error: ! Internal server error, for (POST) [/vat-through-software/direct-debit-yes] ->
  exception=[java.lang.IllegalArgumentException: requirement failed: 'RedirectUrl(/vat-through-software/make-payment/50000/7/2017/2017-07-31/VAT%20Return%20Debit%20Charge/2017-12-07/VAT-RTN-DEBIT-2)' is not a valid continue URL. hence its not applicable for our scenario

PR #981, opened 2025-04-16
[DL-16036] Upgrade for Java 21

Jake-Raffe:


Jake-Raffe:
This should help the bobby warning, will check after merge.
Not going to mess around with the app.routes as part of this PR.

PR #988, opened 2025-08-11
DL-17032: Updated dependency for jackson-core-vulnerability bobby rul…

bwardak:
…e issue

### DIFF
--- a/app/controllers/PaymentHistoryController.scala
+++ b/app/controllers/PaymentHistoryController.scala
@@ -22,6 +22,7 @@ import audit.models.ViewVatPaymentHistoryAuditModel
 import common.SessionKeys
 import config.{AppConfig, ServiceErrorHandler}
 import connectors.httpParsers.ResponseHttpParsers.HttpResult
+
 import javax.inject.{Inject, Singleton}
 import models.viewModels.{PaymentsHistoryModel, PaymentsHistoryViewModel}
 import models.{CustomerInformation, ServiceResponse}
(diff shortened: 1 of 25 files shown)


app/controllers/PaymentHistoryController.scala
```diff
@@ -54,17 +56,16 @@ class PaymentHistoryController @Inject()(paymentsService: PaymentsService,
...
-        serviceInfoContent <- serviceInfoService.getPartial
         migratedWithin15Months = customerMigratedWithin15M(migrationDate)
         payments <- paymentsService.getPaymentsHistory(user.vrn, dateService.now(), migrationDate)
-      } yield {
-        val clientName = request.session.get(SessionKeys.mtdVatvcAgentClientName)
-        val showPreviousPaymentsTab: Boolean = (migratedWithin15Months || migrationDate.isEmpty) && user.hasNonMtdVat
-        generateViewModel(
+        clientName = request.session.get(SessionKeys.mtdVatvcAgentClientName)
+        showPreviousPaymentsTab: Boolean = (migratedWithin15Months || migrationDate.isEmpty) && user.hasNonMtdVat
+        yieldResult <- generateViewModel(
```
bwardak, line 68: Had to move what was in the yield into the for block. Would this big chunk of code need to be refactored?

Jake-Raffe: Not the end of the world. For now the priority is closing the ticket so think it's fine as is

test/views/payments/NoPaymentsViewSpec.scala
```diff
@@ -83,6 +83,7 @@ class NoPaymentsViewSpec extends ViewBaseSpec {
     "render breadcrumbs which" should {
 
       "have the text 'Business tax account'" in {
+        println(document)
```
Jake-Raffe, line 86: println

Jake-Raffe (approved):

PR #1002, opened 2026-01-30
DL-18103: Add company name and VRN to POA schedule page

bwardak:


app/models/viewModels/PaymentsOnAccountViewModel.scala
```diff
@@ -53,7 +53,8 @@ case class PaymentsOnAccountViewModel(
   changedOn: Option[LocalDate],
   currentPeriods: List[VatPeriod],
   pastPeriods: List[VatPeriod],
-  nextPayment: Option[PaymentDetail]
+  nextPayment: Option[PaymentDetail],
+  displayName: Option[String] = None
```
Jake-Raffe: I don't think this needs a default None. It's only used once, with the value passed in so the default could just cause an error in future if someone removes the argument and it wouldn't cause an error

Jake-Raffe: Same for POAController L111

app/controllers/PaymentsOnAccountController.scala
```diff
@@ -38,6 +38,7 @@ class PaymentsOnAccountController @Inject() (
     authorisedController: AuthorisedController,
     dateService: DateService,
     paymentsOnAccountService: PaymentsOnAccountService,
```
Jake-Raffe, line 38: Please could we clean up some of these warnings while we're here and run the auto import clean up

Image

app/controllers/PaymentsOnAccountController.scala
```diff
@@ -58,10 +59,11 @@ class PaymentsOnAccountController @Inject() (
           if (appConfig.features.poaActiveFeatureEnabled()) {
             (for {
               serviceInfoContent <- serviceInfoService.getPartial
+              entityNameResult <- accountDetailsService.getEntityName(user.vrn)
               today = dateService.now()
               standingRequestOpt <- paymentsOnAccountService.getPaymentsOnAccounts(user.vrn)
               obligationsResult <- vatDetailService.getReturnObligations(user.vrn)
-              viewResult <- handleObligationsAndPoa(serviceInfoContent, today, standingRequestOpt, obligationsResult, view)
+              viewResult <- handleObligationsAndPoa(serviceInfoContent, today, standingRequestOpt, obligationsResult, view, entityNameResult.toOption)
```
Jake-Raffe: Can we avoid the double option parameter by doing this instead?

Image

app/views/payments/PaymentsOnAccountView.scala.html
```diff
@@ -50,13 +50,17 @@
...
+        <h1 class="govuk-heading-xl govuk-!-margin-bottom-4">
           @if(user.isAgent) {
             @messages("paymentsOnAccount.agent.heading")
           } else{
             @messages("paymentsOnAccount.heading")
           }
         </h1>
+          <span class="govuk-caption-m">@messages("vatDetails.vrn", user.vrn)</span>
+          @model.displayName.map { entityName =>
+          <span class="govuk-caption-m govuk-!-margin-bottom-4">@entityName</span>
```
Jake-Raffe, line 62: Have we had confirmation that if the call to `entityNameResult <- accountDetailsService.getEntityName` fails, we just turn it into an option and don't display it on the card with the map, rather than pursuing the error?

app/models/viewModels/PaymentsOnAccountViewModel.scala
bwardak: Just to note I've done the same cleanup on the annual accounting side since the code flow was practically identical

app/views/payments/PaymentsOnAccountView.scala.html
Jake-Raffe: Confirmed, is following the agreed logic for vat details page and annual accounting. If the call fails then the vat-overview page would error, meaning they would never access this page anyway.

Jake-Raffe (approved):

